PR #15, opened 2014-06-21
Updates to Library-Dependencies under tutorial

jaceklaskowski:


src/tutorial/08-Library-Dependencies.md
```diff
@@ -131,24 +138,25 @@ the `groupID`), sbt will add your project's Scala version to the artifact
 name. This is just a shortcut. You could write this without the `%%`:
 
 ```scala
-libraryDependencies += "org.scala-tools" % "scala-stm_2.9.1" % "0.3"
+libraryDependencies += "org.scala-tools" % "scala-stm_2.11.1" % "0.3"
```
lrytz: In 2.11, dependencies are cross-versioned against the binary version by default. So I think this should probably be `scala-stm_2.11`.

PR #93, opened 2015-01-03
Docs: demo 'global' lib exclusion.

jotomo:
Add an example of how to exclude a transitive dependency for all dependencies in a sequence.

As I've just learned at http://stackoverflow.com/a/25748761/935239 and thought would be helpful in the reference documentation.

###
Is this well placed in the reference documentation or should it remain on SO?
Note that I didn't build this locally, but used the GitHub editor; but that's what Travis-CI is for, no? :-)

### DIFF
--- a/src/reference/02-DetailTopics/03-Dependency-Management/03-Library-Management.md
+++ b/src/reference/02-DetailTopics/03-Dependency-Management/03-Library-Management.md
@@ -296,6 +296,16 @@ libraryDependencies +=
 
 See [ModuleID](../api/sbt/ModuleID.html) for API details.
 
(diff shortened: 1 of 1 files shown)


typesafehub-validator:
Hi @jotomo,

jotomo:
CLA signed. Travis build failed with an unrelated error (dpkg failed to install a package ...).

eed3si9n:
sbt 0.13.8 will have `excludeDependencies`, which I think should be the way to exclude things going forward. See sbt/sbt#1748.

jotomo:
Cool, _excludeDependencies_ is much cleaner. Thanks @eed3si9n !

PR #80, opened 2015-01-06
Add detailed installation instructions for Ubuntu

lutzh:
http://www.scala-sbt.org/0.13/tutorial/Installing-sbt-on-Linux.html refers to the .deb file, but not all Ubuntu users realize that's what they should use. We should provide instructions rather than just the file, along the lines of https://gist.github.com/visenger/5496675


typesafehub-validator:
Hi @lutzh,

lutzh:
Sigh.. instead of linking the issue and the PR, hub converted the issue.. sorry for that. So I wrote up something, it's only in English, though. How do we usually do the translations? Do I have to provide them to get this merged?

src/tutorial/01-Setup/c.md
```diff
@@ -26,6 +26,20 @@ The following packages are also officially supported:
...
 
+### Ubuntu and other Debian-based Linux Distributions
+
+Linux distributions based on Debian, such as Ubuntu use the DEB format, but usually you don't install your software from a local DEB file. Instead they come with package managers both for the command line (e.g. apt-get, aptitude) or with a graphical user interface (e.g. Synaptic).
+Luckily, they all use the same Advanced Packaging Tool (APT) conventions:
+They will check a number of configured repositories for packages to offer for installation. sbt binaries are published to Bintray, and conveniently Bintray provides an APT repository. You just have to add the repository to the places your package manager will check. So in the directory `/etc/apt/sources.list.d`, create a file `sbt.list` with the contents `deb http://dl.bintray.com/sbt/debian /`
+
+For example like this:
+
+    echo "deb http://dl.bintray.com/sbt/debian /" | sudo tee -a /etc/apt/sources.list.d/sbt.list
```
eed3si9n: Does this require `apt-get update` afterwards? Also I think we should show `apt-get install sbt` above.

eed3si9n:
I personally don't mind hub doing this. I can do the translation and ping the other translators after merge.

src/tutorial/01-Setup/c.md
jsuereth: yes it does.

jsuereth:
Thanks @lutzh !!!

PR #95, opened 2015-01-17
Add sbt-phantomjs to the community plugin list

saturday06:
sbt-phantomjs is an automated installer and configurator for PhantomJS


eed3si9n:
Thanks

PR #102, opened 2015-02-11
Use https to fetch packages

benmccann:


eed3si9n:
I pushed the change, and I can see the change - http://www.scala-sbt.org/0.13/tutorial/Installing-sbt-on-Linux.html

benmccann:
Thanks!

eed3si9n:
#127 says https causes segfault.

PR #107, opened 2015-02-16
Update sbt 0.12 => 0.13 when using cs install sbt

netroby:
Update sbt 0.12 => 0.13 when using cs install sbt

### DIFF
--- a/src/reference/02-DetailTopics/01-Using-sbt/06-Scripts.md
+++ b/src/reference/02-DetailTopics/01-Using-sbt/06-Scripts.md
@@ -30,7 +30,7 @@ mode that only requires a JRE installed.
 Install [conscript](https://github.com/n8han/conscript).
 
 ```
-\$ cs sbt/sbt --branch 0.13.0
+\$ cs sbt/sbt --branch $app_version$
 ```
 
 This will create two scripts: `screpl` and `scalas`.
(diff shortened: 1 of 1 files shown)


eed3si9n:
A better change would be to replace it with `$app_version$` so Pamflet would replace it with the latest.

netroby:
Sorry, the document is a bit out of date. so i want to let it go forward.

the 0.13 branch is stable and can be safe used now.
so the little change will be ok.

I will keep my eyes on the document , update the document if you like me do this.

If you have better idea, please let me know.

src/reference/02-DetailTopics/01-Using-sbt/06-Scripts.md
```diff
@@ -30,7 +30,7 @@ mode that only requires a JRE installed.
 Install [conscript](https://github.com/n8han/conscript).
 
 ```
-\$ cs sbt/sbt --branch 0.12.0
+\$ cs sbt/sbt --branch 0.13.0
```
eed3si9n: This is using `0.13.0` branch instead of the latest `0.13`.
I think the best way is to say `$app_version$` there, which the documentation engine would substitute to 0.13.7 from https://github.com/sbt/website/blob/master/src/reference/template.properties

netroby:
Hello, i pushed as you told me . can you check it again for me ? is it ok?

eed3si9n:
LGTM. Thanks for the contribution.

netroby:
Thanks for you guide me .

netroby:
どうもありがとうございました

eed3si9n:
别客气

PR #114, opened 2015-02-23
Code sample fixes

fernandezpablo85:
The text 

> The key constructors have two string parameters

Matches neither the code sample nor the referenced `sbt.Keys.scala` file

### DIFF
--- a/src/tutorial/11-Custom-Settings.md
+++ b/src/tutorial/11-Custom-Settings.md
@@ -32,12 +32,12 @@ Keys have one of three types. `SettingKey` and `TaskKey` are described in
 Some examples from [Keys](../sxr/sbt/Keys.scala.html):
 
 ```scala
-val scalaVersion = settingKey[String]("The version of Scala used for building.")
-val clean = taskKey[Unit]("Deletes files produced by the build, such as generated sources, compiled classes, and task caches.")
+val scalaVersion = settingKey[String]("scala-version", "The version of Scala used for building.")
+val clean = taskKey[Unit]("clean", "Deletes files produced by the build, such as generated sources, compiled classes, and task caches.")
 ```
 
 The key constructors have two string parameters: the name of the key
-(`"scalaVersion"`) and a documentation string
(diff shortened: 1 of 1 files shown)


src/tutorial/11-Custom-Settings.md
```diff
@@ -32,12 +32,12 @@ Keys have one of three types. `SettingKey` and `TaskKey` are described in
...
 ```scala
-val scalaVersion = settingKey[String]("The version of Scala used for building.")
-val clean = taskKey[Unit]("Deletes files produced by the build, such as generated sources, compiled classes, and task caches.")
+val scalaVersion = settingKey[String]("scala-version", "The version of Scala used for building.")
+val clean = taskKey[Unit]("clean", "Deletes files produced by the build, such as generated sources, compiled classes, and task caches.")
 ```
 
 The key constructors have two string parameters: the name of the key
-(`"scalaVersion"`) and a documentation string
+(`"scala-version"`) and a documentation string
```
jsuereth: So, in sbt 0.13 we create a "backwards compatible mode" for camel-case vs. '-'.  IN all of sbt, the string `scala-version` CAN be used, but wont' show up in help documentation.   Yes the code still has the backwards compatible name.  I think perhaps we need to clarify.

src/tutorial/11-Custom-Settings.md
```diff
@@ -32,12 +32,12 @@ Keys have one of three types. `SettingKey` and `TaskKey` are described in
 Some examples from [Keys](../sxr/sbt/Keys.scala.html):
 
 ```scala
-val scalaVersion = settingKey[String]("The version of Scala used for building.")
-val clean = taskKey[Unit]("Deletes files produced by the build, such as generated sources, compiled classes, and task caches.")
+val scalaVersion = settingKey[String]("scala-version", "The version of Scala used for building.")
```
jsuereth: The previous code is still the *conceptual* constructor.  We would have it be exactly that if we weren't maintaining compatibility w/ 0.12 build scripts.

PR #118, opened 2015-03-02
Adding link to test reporter plugin

fleipold:
The tabular test reporter produces test results
well suited for further processing and analysis.


eed3si9n:
Thanks

PR #138, opened 2015-04-14
Integrate notes on incremental compilation

gkossakowski:
This PR integrates my notes on incremental compilation that live here: https://github.com/gkossakowski/sbt/wiki/Incremental-compiler-notes

After copy&pasting information about implementation details of inc and name hashing I went on to update it with latest developments. My notes were written before name hashing got merged into sbt.

The overall shape of the `07-Incremental-Recompilation.md` is slightly better with this PR than without it. I recognize it's not ideal but I've ran out of time to improve it further.


eed3si9n:
Thanks for looking into this.

Issue #106, opened 2015-02-14
version switch


eed3si9n:
Any improvement on this front would be great
## problem 1

When you get to 0.13.5 from Google for example http://www.scala-sbt.org/0.13.5/docs/Getting-Started/Welcome.html, 0.13 isn't on the version switch menu.
0.13 should be at the top, and it should navigate to http://www.scala-sbt.org/0.13/tutorial/ if you're on Getting Started guide; http://www.scala-sbt.org/0.13/docs/ otherwise.
## problem 2

If you can jump to the corresponding page on 0.13 that'll be cool. I think there are some redirects that I generated.
## problem 3

When you are on 0.13 tutorial or reference, version switch control is not displayed.
Again, I think it's ok for the first round to jump to the index page.
## notes

Here are the stuff to fix
- http://www.scala-sbt.org/versions.js
- http://www.scala-sbt.org/0.13.2/docs/_static/set-versions.js

metasim:
What updates the `version.js` file? Was that a feature of Sphinx, or a part of the publish process?

eed3si9n:
This seems to be it:
https://github.com/sbt/sbt/blob/v0.13.5/project/Docs.scala#L98
I'm ok with manually updating this file since we now use 0.13 for all 0.13.x.

metasim:
Am I right that the old docs are part of the sbt/sbt repo? So problem 1 will be address by code in there?

eed3si9n:
The raw files are here, so I was thinking about fixing it there - https://github.com/sbt/sbt.github.com

dwijnand:
Tweaked in https://github.com/sbt/website/commit/dd1570a18adb27300fa6d227a63598fffc8809b0 and the commits in that neighborhood.

PR #145, opened 2015-04-23
Integration of version switch from previous doc system.

metasim:
To partially address #106.

This is probably just a draft attempt, at least from a visual standpoint. I'm not super great at CSS, so I probably violated some visual design best practice, but it's good enough for someone to improve upon.

Needs to be tested against a full site publish to ensure cross version switching works as expected.


eed3si9n:
I was able to confirm the selector show up on the latest doc. Let's merge it up and see how it works :)

PR #147, opened 2015-04-25
Fix index page of Japanese tutorial

seratch:
Sorry for bothering you again. I just fixed some issues [here](http://www.scala-sbt.org/0.13/tutorial/ja/index.html).


eed3si9n:
Please come by as often as you want.

PR #170, opened 2015-07-15
Add sbt-api-mappings

Atry:
sbt-api-mappings is a Sbt plugin that fills apiMappings for common Scala libraries


eed3si9n:
Thanks!

PR #174, opened 2015-07-22
added sbt-planout4j plugin

reikje:


eed3si9n:
Thanks!

reikje:
Looks like Travis didn't like it :)

eed3si9n:
It's currently a manual process to push a new site. I should really automate that using Jenkins.

PR #189, opened 2015-08-25
coding guideline

eed3si9n:
Draft guideline on sbt 1.0 refactoring.

/review @jsuereth, @dwijnand, @Duhemm

### DIFF
--- a/src/reference/03-Developers-Guide/02-Coding-Guideline/00.md
+++ b/src/reference/03-Developers-Guide/02-Coding-Guideline/00.md
@@ -165,3 +165,44 @@ sbt-houserules comes with scalariform.
 #### Avoid procedure syntax
 
 Return `Unit`.
+
+#### Typeclass instances are encouraged to be defined in the companions
+
+```scala
+final class FooID {}
+object FooID {
+  implicit val fooIdPicklerUnpicker: PicklerUnpickler[FooID] = ???
+}
+```
+
+#### Implicit converter for syntax (enrich-my-library) should be imported
+
+Avoid defining implicit converters in companion objects and package objects.
+
+Suppose IO module introduces `URL` enrichment called `RichURI`;
+and LibraryManagement introduces `String` enrichment called `GroupID` (for `ModuleID` syntax).
+These implicit converters should be defined in a trait named `IOSyntax` and `LibraryManagementSyntax` respectively:
+
+```scala
+package sbt.io
+
+trait IOSyntax {
+  implicit def uriToRichURI(uri: URI): RichURI = new RichURI(uri)
+}
+```
+
+Next each layer should define an object named `syntax`:
+
+```scala
+package sbt.io
+
+object syntax extends IOSyntax
+```
+
+When all the layers are available, `sbt` package should also define an object called `syntax`:
+
+```scala
+package sbt
+
+object syntax extends IOSyntax with LibraryManagementSyntax with ...
+```
--- a/src/reference/03-Developers-Guide/02-Coding-Guideline/00.md
+++ b/src/reference/03-Developers-Guide/02-Coding-Guideline/00.md
@@ -139,7 +139,7 @@ Default to `Vector`. Use `List` if constant prepending is needed.
 Use `Array` if Java interop is needed.
 Note using mutable collection is perfectly fine within the implementation.
 
-#### Avoid `Set` in the signature
+#### Avoid calling `toSeq` or anything side-effecty on `Set`
 
 `Set` is fine if you stick with set operations, like `contains` and `subsetOf`.
 More often than not, `toSeq` is called explicitly or implicitly,
--- a/src/reference/03-Developers-Guide/02-Coding-Guideline/00.md
+++ b/src/reference/03-Developers-Guide/02-Coding-Guideline/00.md
@@ -181,28 +181,24 @@ Avoid defining implicit converters in companion objects and package objects.
 
 Suppose IO module introduces `URL` enrichment called `RichURI`;
 and LibraryManagement introduces `String` enrichment called `GroupID` (for `ModuleID` syntax).
-These implicit converters should be defined in a trait named `IOSyntax` and `LibraryManagementSyntax` respectively:
+These implicit converters should be defined in an object named `syntax` in respective packge:
 
 ```scala
 package sbt.io
 
-trait IOSyntax {
+object syntax {
   implicit def uriToRichURI(uri: URI): RichURI = new RichURI(uri)
 }
 ```
 
-Next each layer should define an object named `syntax`:
-
-```scala
-package sbt.io
-
-object syntax extends IOSyntax
-```
-
-When all the layers are available, `sbt` package should also define an object called `syntax`:
+When all the layers are available, `sbt` package should also define an object called `syntax`,
+which forwards all the implicit converters from all the layers:
 
 ```scala
 package sbt
 
-object syntax extends IOSyntax with LibraryManagementSyntax with ...
+object syntax {
+  implicit def uriToRichURI(uri: URI): io.RichURI = io.syntax.uriToRichURI(uri)
+  ....
+}
 ```
--- a/src/reference/03-Developers-Guide/02-Coding-Guideline/00.md
+++ b/src/reference/03-Developers-Guide/02-Coding-Guideline/00.md
@@ -73,6 +73,11 @@ A module may be declared internal if there's no use for public.
 
 The `-Xfatal-warnings` may be removed if there are unavoidable warnings.
 
+#### Package name and organization name
+
+Use the package name appended with the layer name, such as `sbt.io` for IO layer.
+The organization name for the publish artifacts should remain `org.scala-sbt`.
+
 ### Binary resiliency
 
 A good overview on the topic of binary resiliency is [Josh's 2012][jsuereth2012] talk
--- a/src/reference/03-Developers-Guide/02-Coding-Guideline/00.md
+++ b/src/reference/03-Developers-Guide/02-Coding-Guideline/00.md
@@ -52,6 +52,27 @@ Avoid exposing external classes out to API, except for standard Scala and Java c
 
 A module may be declared internal if there's no use for public.
 
+#### Compiler flags
+
+```
+-encoding utf8
+-deprecation
+-feature
+-unchecked
+-Xlint
+-language:higherKinds
+-language:implicitConversions
+-Xfuture
+-Yinline-warnings
+-Yno-adapted-args
+-Ywarn-dead-code
+-Ywarn-numeric-widen
+-Ywarn-value-discard
+-Xfatal-warnings
+```
+
+The `-Xfatal-warnings` may be removed if there are unavoidable warnings.
+
 ### Binary resiliency
 
 A good overview on the topic of binary resiliency is [Josh's 2012][jsuereth2012] talk
@@ -90,7 +111,7 @@ case class involves various codegen, which makes it harder to maintain binary co
 
 #### Prefer method overloading over default parameter values
 
-The default parameter values are effectivly codegen,
+The default parameter values are effectively codegen,
 which makes it difficult to maintain.
 
 ### Other public API matter
@@ -106,7 +127,7 @@ Define datatypes.
 `def apply` should be reserved for factory methods
 in a companion object that returns type `T`.
 
-#### Use speicific datatype (`Vector`, `List`, or `Array`) rather than `Seq`.
+#### Use specific datatype (`Vector`, `List`, or `Array`) rather than `Seq`.
 
 `scala.Seq` is `scala.collection.Seq`, which is not immutable.
 Default to `Vector`. Use `List` if constant prepending is needed.
@@ -135,3 +156,7 @@ incremental compiler.
 #### Use scalariform
 
 sbt-houserules comes with scalariform.
+
+#### Avoid procedure syntax
+
+Return `Unit`.
--- a/src/reference/03-Developers-Guide/02-Coding-Guideline/00.md
+++ b/src/reference/03-Developers-Guide/02-Coding-Guideline/00.md
@@ -1,0 +1,137 @@
+---
+out: Coding-Guideline.html
+---
+
+  [jsuereth2012]: http://jsuereth.com/scala/2012/04/22/scaladays.html
+  [pins_trait127]: http://www.artima.com/pins1ed/traits.html#12.7
+  [mima]: https://github.com/typesafehub/migration-manager
+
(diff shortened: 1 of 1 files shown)


src/reference/03-Developers-Guide/02-Coding-Guideline/00.md
```diff
@@ -0,0 +1,137 @@
...
+
+Typeclass pattern with pure trait might be easier to maintain binary compatibility than subclassing.
+
+#### Avoid case classes, use sbt-datatype
+
+case class involves various codegen, which makes it harder to maintain binary compatibility over time.
+
+#### Prefer method overloading over default parameter values
+
+The default parameter values are effectivly codegen,
```
dwijnand: effectively

src/reference/03-Developers-Guide/02-Coding-Guideline/00.md
```diff
@@ -0,0 +1,137 @@
...
+We will cross build against Scala 2.10.
+
+#### Clean up old deprecation
+
+Before 1.0 is release, we should clean up deprecations.
+
+#### Aim for zero warnings (except deprecation)
+
+On Scala 2.11 we should aim for zero warnings.
+One exception may be deprecation if it's required for 2.10 cross building.
```
dwijnand: even then we could (should) create skims in cross-version sources imo

src/reference/03-Developers-Guide/02-Coding-Guideline/00.md
```diff
@@ -0,0 +1,137 @@
...
+
+A good overview on the topic of binary resiliency is [Josh's 2012][jsuereth2012] talk
+Binary resiliency.
+The guideline here applies mostly to publicly exposed APIs.
+
+#### MiMa
+
+Use [MiMa][mima].
+
+#### Public traits should contain `def` declarations only
```
dwijnand: Does this mean only abstract/unimplemented? or also implemented defs?

src/reference/03-Developers-Guide/02-Coding-Guideline/00.md
```diff
@@ -0,0 +1,137 @@
...
+#### Avoid Stringly-typed programming
+
+Define datatypes.
+
+#### Avoid overuse of `def apply`
+
+`def apply` should be reserved for factory methods
+in a companion object that returns type `T`.
+
+#### Use speicific datatype (`Vector`, `List`, or `Array`) rather than `Seq`.
```
dwijnand: specific

src/reference/03-Developers-Guide/02-Coding-Guideline/00.md
```diff
@@ -0,0 +1,137 @@
...
+#### Prefer method overloading over default parameter values
+
+The default parameter values are effectivly codegen,
+which makes it difficult to maintain.
+
+### Other public API matter
+
+Here are other guidelines about public API.
+
+#### Avoid Stringly-typed programming
```
dwijnand: I would argue for the inclusion of "Avoid boolean blindness"

src/reference/03-Developers-Guide/02-Coding-Guideline/00.md
```diff
@@ -0,0 +1,137 @@
...
+
+#### Finalize the leaf classes
+
+If there's no need to keep the classes open, finalize it.
+
+#### Typeclass and subclass inheritance
+
+Typeclass pattern with pure trait might be easier to maintain binary compatibility than subclassing.
+
+#### Avoid case classes, use sbt-datatype
```
dwijnand: Are case objects ok?

src/reference/03-Developers-Guide/02-Coding-Guideline/00.md
jsuereth: If practical...

src/reference/03-Developers-Guide/02-Coding-Guideline/00.md
jsuereth: Yeah

src/reference/03-Developers-Guide/02-Coding-Guideline/00.md
```diff
@@ -0,0 +1,137 @@
...
+in a companion object that returns type `T`.
+
+#### Use speicific datatype (`Vector`, `List`, or `Array`) rather than `Seq`.
+
+`scala.Seq` is `scala.collection.Seq`, which is not immutable.
+Default to `Vector`. Use `List` if constant prepending is needed.
+Use `Array` if Java interop is needed.
+Note using mutable collection is perfectly fine within the implementation.
+
+#### Avoid `Set` in the signature
```
jsuereth: I still think this is overkill.  If you're going to remove set, then provide something they should use instead, like `collection.mutable.LinkedHashSet`.   I think relying on ordering for Iterable/Set is generally evil, so I think you're putting the wrong lesson from those mistakes in here.

src/reference/03-Developers-Guide/02-Coding-Guideline/00.md
```diff
@@ -0,0 +1,137 @@
...
+
+`Set` is fine if you stick with set operations, like `contains` and `subsetOf`.
+More often than not, `toSeq` is called explicitly or implicitly,
+or some side effecting method is called from `map`.
+This introduces non-determinism to the code.
+
+#### Avoid calling `toSeq` on `Map`
```
jsuereth: This is the lesson we should have.  Additionally "mkString" and many others.

src/reference/03-Developers-Guide/02-Coding-Guideline/00.md
eed3si9n: "declaration" is signature only, without implementation.

src/reference/03-Developers-Guide/02-Coding-Guideline/00.md
jsuereth: +1.  Also FYI - we kind of do stringly typed a lot in sbt, we just hide it behind "AttributeMap" :)

Duhemm:
I think this guide should mention the recommended compiler options. How about `-feature`, `-unchecked` and `-deprecation`?

eed3si9n:
@Duhemm Ideally `-Xfatal-warnings` too. I'm ok with `-language:_`.

Duhemm:
`-language:_` means enable all language features, right?

dwijnand:
Yes.

Personally I would recommend:

```
-encoding utf8
-deprecation
-feature
-unchecked
-Xlint
-language:higherKinds
-language:implicitConversions
-language:postfixOps
-Xfuture
-Yinline-warnings
-Yno-adapted-args
-Ywarn-dead-code
-Ywarn-numeric-widen
-Ywarn-value-discard
```

There's also `-Ywarn-unused-import` which is good but it's both 2.11 only and mustn't be present in console (especially with fatal warning).

I've personally not enabled all language flags as I'm not sure about them all, particularly I don't use structural types enough to recognise when I'm using them (and paying the runtime price), but I'm happy to go with `-language:_`.

src/reference/03-Developers-Guide/02-Coding-Guideline/00.md
eed3si9n: What's the alternative to Boolean blindness? Always supply parameter name at call site, or define data types always?

dwijnand: I think it's create your own descriptive sealed hierachies. So there's less enable=true vs disable=true confusion.

src/reference/03-Developers-Guide/02-Coding-Guideline/00.md
eed3si9n: I'm not banning `Set` completely, I am suggest we avoid passing `Set` around in API.
Choosing `Set` often is an early optimization to promise less about the API, which is laudable, but `Set` as a type cannot prevent the non-determism to leak out. And let's face that fact that we are talking about build tool. Most of the function will result to one side effect or another like things getting compiles or JARs getting downloaded from the Internet.

As an alternative I would suggest an interface that define `def isMemberOfFoo(x: X): Boolean` or a `Vector`.

src/reference/03-Developers-Guide/02-Coding-Guideline/00.md
eed3si9n: `IO` has ` def touch(file: File, setModified: Boolean = true): Unit`. How would you fix it?

dwijnand: Perhaps

```scala
sealed trait SetModified
case object TouchSetsModified extends SetModified
case object TouchLeavesModifies extends SetModified
```

But I'm still working on better alternatives.

Perhaps this could be a lax guideline. Essentially I think we should avoid unclear parameters (`disable = false`), multiple booleans and unclear method calls (eg. `touch(file, true, false, true)`).

eed3si9n:
>     -language:postfixOps

I am actually ok removing postfixOps. I can't think of good of it besides some Englishy DSLs.

dwijnand:
I like postfixOps because it works well with infix notation, which perhaps we should touch on...

```scala
xs map f filter p toVector
```

src/reference/03-Developers-Guide/02-Coding-Guideline/00.md
eed3si9n: So why don't we say that anything that needs default value should likely turn into a pseudo case class:

```scala
case class TouchOptions(setModified: Boolean)
object TouchOptions {
  def default: TouchOptions = TouchOptions(setModified = true)
}
```

and say that we always use `true` to mean yes/enable, and `false` to mean no/disable.

```scala
def touch(file: File): Unit = touch(file, TouchOption.default)
```

dwijnand: I'm not convinced..

Perhaps instead `IO.touch(file: File)` should just do whatever `setModified = true` does, and there should be a `IO.touchIgnoreModified(file: File)` for the alternative.

eed3si9n:
I would normally go for infix notation if it feel like an operator like collection library methods, but avoid postfix.
In the example you gave, I'd probably put the infix part in paren:

```scala
(xs map f filter p).toVector
```

src/reference/03-Developers-Guide/02-Coding-Guideline/00.md
eed3si9n: Do we create 2^n variants for methods that takes multiple of these flags?

```
def copy(sources: Traversable[(File, File)], overwrite: Boolean = false, preserveLastModified: Boolean = false): Set[File]
```

One convenience with using plain `Boolean` is that the compilation error can tell you what values you can pass in.

dwijnand: No, probably not 2^n variants. As I say, perhaps it should be "consider avoiding using booleans".

Hypothetically for your example I might say we should drop the method variant which takes a Traversable just to map it (pushing that to the user), as well as drop it optionally (via the boolean) calling `copyLastModified` (again, pushing that to the user). Which would leave perhaps:

```scala
/**
  * Copies the contents of the first file (the source) to the location of the second file (the target), as long as the target file is either missing or older than the source file according to last modified times. If the source is a directory the corresponding directory is created. Any parent directories that do not exist are created. The target file is returned, whether or not it was updated by this method.
*/
def copy(source: File, target: File): File

/**
  * Copies the contents of the first file (the source) to the location of the second file (the target). If the source is a directory the corresponding directory is created. Any parent directories that do not exist are created. The target file is then returned.
*/
def copyForce(source: File, target: File): File
```

jsuereth: I wouldn't remove the methods that are the ones we mostly use (like copy of Traversable).

eed3si9n:
Updated the guide. I'm ok with removing any contentious parts out of this PR, so we can move forward.

dwijnand:
I would still prefer something like "consider avoiding using booleans", but I'm happy to merge this as is. :+1: Good work.

src/reference/03-Developers-Guide/02-Coding-Guideline/00.md
```diff
@@ -0,0 +1,204 @@
...
+Before 1.0 is release, we should clean up deprecations.
+
+#### Aim for zero warnings (except deprecation)
+
+On Scala 2.11 we should aim for zero warnings.
+One exception may be deprecation if it's required for 2.10 cross building.
+
+### Modular design
+
+#### Aim small
```
jsuereth: Aim small, Aim flexible 

Flexible methods => less exposed.

src/reference/03-Developers-Guide/02-Coding-Guideline/00.md
```diff
@@ -0,0 +1,204 @@
...
+
+The fewer methods we can expose to the build user, the easier it becomes to maintain it.
+
+#### Public APIs should be coded against "interfaces"
+
+Code against interface.
```
jsuereth: After this PR, I'd like to try to add some reasoning behind all this.

src/reference/03-Developers-Guide/02-Coding-Guideline/00.md
```diff
@@ -0,0 +1,204 @@
...
+
+Making independent modules with fewer dependent libraries make it easier to reuse them.
+
+#### Hide external classes
+
+Avoid exposing external classes out to API, except for standard Scala and Java classes.
+
+#### Hide internal modules
+
+A module may be declared internal if there's no use for public.
```
jsuereth: We should note: that you should not depend on another module's internals (unless you must).

src/reference/03-Developers-Guide/02-Coding-Guideline/00.md
```diff
@@ -0,0 +1,204 @@
...
+
+`scala.Seq` is `scala.collection.Seq`, which is not immutable.
+Default to `Vector`. Use `List` if constant prepending is needed.
+Use `Array` if Java interop is needed.
+Note using mutable collection is perfectly fine within the implementation.
+
+#### Avoid calling `toSeq` or anything side-effecty on `Set`
+
+`Set` is fine if you stick with set operations, like `contains` and `subsetOf`.
+More often than not, `toSeq` is called explicitly or implicitly,
```
jsuereth: I bet we could make an abide rule to detect this and throw an exception any time a `Set` appears in an expression of type `Seq` (or you see a toSeq)

jsuereth:
LGTM, good work on this, we'll have to evovle it as we go.   Should I worry about build failures or just merge?

eed3si9n:
Build failure is probably about Travis not downloading Latex.

PR #190, opened 2015-09-05
Reverse nest internal packages in the coding guideline.

dwijnand:
Given the way Scala nests things, it makes more sense for internal things to see
package x specific things.

### DIFF
--- a/src/reference/03-Developers-Guide/02-Coding-Guideline/00.md
+++ b/src/reference/03-Developers-Guide/02-Coding-Guideline/00.md
@@ -37,7 +37,7 @@ Code against interface.
 
 #### Hide implementation details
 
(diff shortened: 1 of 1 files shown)


src/reference/03-Developers-Guide/02-Coding-Guideline/00.md
```diff
@@ -37,7 +37,7 @@ Code against interface.
 
 #### Hide implementation details
 
-The implementation details should be hidden behind `sbt.internal.x` package,
+The implementation details should be hidden behind `sbt.x.internal` package,
```
eed3si9n: See my comment on https://github.com/sbt/util/pull/13#issuecomment-137935175

PR #192, opened 2015-09-10
Fix documentation about how to load manifst from a file

sschaef:
The previous code snippet did not compile. We have to specify the file
that needs to be loaded.


eed3si9n:
Thanks!

PR #203, opened 2015-10-19
Minor grammar fix

esamson:
Sorry, I couldn't help it but either "need to *be* explicitly enabled" or  "need to explicitly *be* enabled" sounds better in my head.


eed3si9n:
Thanks

PR #206, opened 2015-11-12
add a bit of doc on runMain

SethTisue:
this is touched on in a few other places in the doc, but
it seemed strange to me that it wasn't covered here


dwijnand:
:+1:

PR #216, opened 2016-01-14
Snippet showing how to invoke specific tasks

benhutchison:
derived from: http://stackoverflow.com/questions/9856204/sbt-skip-tests-in-subproject-unless-running-from-within-that-project

### DIFF
--- a/src/reference/04-Howto/13-Howto-Custom-Tasks.md
+++ b/src/reference/04-Howto/13-Howto-Custom-Tasks.md
@@ -3,13 +3,20 @@ Define Custom Tasks
 
 ### Define a Task that runs tests in specific sub-projects
 
-Defines a task `myTestTask` that will run the `test` Task in specific subprojects  `abc` and `def` but no others:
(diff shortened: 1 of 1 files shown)


typesafehub-validator:
Hi @benhutchison,

src/reference/04-Howto/13-Howto-Custom-Tasks.md
```diff
@@ -0,0 +1,15 @@
+Define Custom Tasks 
+-------------------
+
+### Define a Task that runs tests in specific sub-projects
+
+Defines a task `myTestTask` that will run the `test` Task in specific subprojects  `abc` and `def` but no others:
```
eed3si9n: `def` is a keyword, so it's probably not a good idea to use it here. The current recommendation is to use multi-project build.sbt style. Also wouldn't it be a better example if it's self-complete? i.e. define the subprojects it's referring to?

benhutchison: good points, addressed by another commit

eed3si9n:
Thanks!

PR #225, opened 2016-02-19
added how to guide on sequencing tasks

eed3si9n:


src/reference/04-Howto/20-Howto-Sequencing/00.md
```diff
@@ -0,0 +1,38 @@
...
+This is more constrained compared to the imperative style plain Scala code with side effects such as the follows:
+
+```scala
+def foo(): Unit = {
+  doX()
+  doY()
+}
+```
+
+The benefit of the dependency-oriented programming model is that sbt's task engine is able to reorder the task execution. When possible we run dependent tasks in parallel. Another benefit is that we can deduplicate the graph, and make sure that the `compile` is called once per command execution.
```
dwijnand: do you mean `compile` is special-cased in the task engine?

dwijnand:
LGTM!

I have one question, but mostly for my own benefit, not a blocker.

src/reference/04-Howto/20-Howto-Sequencing/04-Howto-Dynamic-Input-Task.md
```diff
@@ -0,0 +1,61 @@
...
+  settings(
+    run in Compile := (Def.inputTaskDyn {
+      import sbt.complete.Parsers.spaceDelimited
+      val args = spaceDelimited("<args>").parsed
+      Def.taskDyn {
+        (actualRun in Compile).toTask(" " + args.mkString(" ")).value
+        openbrowser
+      }
+    }).evaluated,
+    actualRun in Compile := Defaults.runTask(
```
smarter: It'd be helpful if this tutorial explained how you came up with the definition of `actualRun in Compile`.

src/reference/04-Howto/20-Howto-Sequencing/00.md
eed3si9n: no all tasks are deduplicated.

src/reference/04-Howto/20-Howto-Sequencing/00.md
```diff
@@ -0,0 +1,38 @@
...
+This is more constrained compared to the imperative style plain Scala code with side effects such as the follows:
+
+```scala
+def foo(): Unit = {
+  doX()
+  doY()
+}
+```
+
+The benefit of the dependency-oriented programming model is that sbt's task engine is able to reorder the task execution. When possible we run dependent tasks in parallel. Another benefit is that we can deduplicate the graph, and make sure that the task evaluation, such as `compile in Compile`, is called once per command execution, as oppose to compiing the same source many times.
```
dwijnand: s/compiing/compiling/

dwijnand: s/oppose/opposed/

PR #227, opened 2016-03-01
Use after_script and "-delete" to clean up files

graingert:
`find | xargs rm` is dangerous (you need `find -print0 | xargs -0 rm`) and slow.

See https://www.gnu.org/software/findutils/manual/html_node/find_html/Deleting-Files.html for a discussion


dwijnand:
I agree.

Could you:
* use `before_cache` instead of `after_script` (I believe it's more correct), and
* add a `-print` to `find` (I think it's good to have in the logs)

Other than those :+1:

graingert:
@dwijnand not sure it needs the logging

PR #230, opened 2016-03-12
Make a pass at slaying typos in Developer Guide

ches:
I caught several typos near the start of reading the Dev Guide, so, well, I cloned and read the rest of it in my editor and fixed things along the way. Some spelling errors, some minor grammar clean-up, some Markdown formatting that makes reading easier. There are two or three spots where I wasn't sure what the author intended to say so I marked them `TODO` or `FIXME`…

Sorry it's kind of a lot to review, but I figured a bunch of smaller pull requests would be no less tedious.


src/reference/90-Developers-Guide/00.md
```diff
@@ -48,13 +48,13 @@ source compatibility of the build during 1.x.y.
...
-The stability here helps maintaining sbt plug ecosystem.
+The stability here helps maintain the sbt plugin ecosystem.
 Our goal for sbt 1.0 is to adopt Semantic Versioning, and maintain
 binary compatibility of the build during 1.x.y.
 
 From the development perspective, maintaining binary compatibility becomes
 an additional constraint that we need to worry about whenever we make changes.
-The of the problem is that sbt 0.13 does not distinguish between public API
-and internal impelmentation. Most things are open to plugins.
+The [what?] of the problem is that sbt 0.13 does not distinguish between public API
```
dwijnand: Perhaps: the meat of the problem? But I'd rather say something like: A major problem is that...

src/reference/90-Developers-Guide/02-Coding-Guideline/00.md
```diff
@@ -90,83 +90,87 @@ Use [MiMa][mima].
 
 #### Public traits should contain `def` declarations only
 
+TODO: "results to codegen at subclass" is unclear
```
dwijnand: True.. I'm not sure what to say instead though..

src/reference/90-Developers-Guide/09-Launcher/01-Launcher-Getting-Started.md
```diff
@@ -116,11 +122,14 @@ interface class will be provided by the launcher, so it is only a
 compile-time dependency. If you are building with sbt, your dependency
 definition would be:
 
-Make the entry point to your class implement 'xsbti.AppMain'. An example
+FIXME
```
dwijnand: (commenting so we don't merge this as is)

src/reference/90-Developers-Guide/09-Launcher/01-Launcher-Getting-Started.md
```diff
@@ -154,7 +163,9 @@ class Main extends xsbti.AppMain
 Next, define a configuration file for the launcher. For the above class,
 it might look like:
 
-Then, `publishLocal` or `+publishLocal` the application to make it
+FIXME
```
dwijnand: (commenting so we don't merge this as is)

dwijnand:
Reviewed. Thanks for for the contribution @ches!

Few things left to do, but LGTM overall.

src/reference/90-Developers-Guide/02-Coding-Guideline/00.md
ches: Thanks for the review Dale!

I need to understand the technical point being made better in order to suggest improved English, but if the intended meaning of "results to codegen" is close to "causes code generation", then I would typically say "results *in*", not "results to" (or "causes").

I *think* I would write something like:

> `val`s or `var`s defined in `trait`s result in code generation when extended, and artificial `Foo\$class.\$init\$` compilation units.

I'm not certain "compilation unit" is the correct term.

Also, it might help to distinguish clearly how `def` differs—is the point that `val` and `var` produce synthesized accessor/modifier property methods for their underlying fields?

src/reference/90-Developers-Guide/00.md
ches: Yep I assumed "meat" or "core" or something like that was the omitted word, but I also agree that it could probably be reworded better.

src/reference/90-Developers-Guide/09-Launcher/01-Launcher-Getting-Started.md
ches: I could probably work out the needed example myself from the context, but it might be easier for someone more familiar with using launcher applications (@eed3si9n or @n8han? :smile:).

src/reference/90-Developers-Guide/02-Coding-Guideline/00.md
dwijnand: The intent is that vals/vars/lazy vals are: (1) problematic for trait lineralisation, (2) problematic for binary compatibility.

But I'm not sure what exactly are and how to exactly explain the specific technical reasons behind it.

src/reference/90-Developers-Guide/09-Launcher/01-Launcher-Getting-Started.md
ches: I pulled this out of [the 0.12 docs][1] where it must have been lost in translation somehow, and updated the version:

```scala
libraryDependencies += "org.scala-sbt" % "launcher-interface" % "1.0.0" % "provided"

resolvers <+= sbtResolver
```

Assuming that is still correct, I'll commit it to this branch.

[1]: http://www.scala-sbt.org/0.12.1/docs/Detailed-Topics/Launcher.html#creating-a-launched-application

src/reference/90-Developers-Guide/09-Launcher/01-Launcher-Getting-Started.md
ches: Okay ditto, this was also in the 0.12 docs and lost somehow:

```
[scala]
  version: 2.9.2
[app]
  org: org.scala-sbt
  name: xsbt-test
  version: 0.12.0
  class: xsbt.test.Main
  cross-versioned: true
[repositories]
  local
  maven-central
[boot]
 directory: ${user.home}/.myapp/boot
```

The preceding code example needs updating to depict a current Scala version there. I assume that `0.12.0` is an arbitrary version number for the application itself, not actually an sbt version as it sort of suggests?

I'll add this back on the branch.

dwijnand: Ah yeah. I think this was machine-converted.

And yeah while we're at it we should move away from org.scala-sbt/xsbt-test/0.12.0/xsbt.test.Main, and use a more acme/foo type example. Good call.

src/reference/90-Developers-Guide/00.md
eed3si9n: I am guessing the *root* of the problem?

src/reference/90-Developers-Guide/01-Modularization/01-Module-Summary.md
```diff
@@ -33,21 +33,21 @@ The responsibility of the serialization API is to turn values into JSON.
 
 #### Util APIs ([sbt/util][utilrepo])
 
-Util APIs provide commonly used features like logging and internal datatypes used by sbt.
+Util APIs provide commonly used features like logging and internal data types used by sbt.
```
eed3si9n: I like my datatype spelled as one word. See also http://www.cs.ox.ac.uk/jeremy.gibbons/publications/dgp.pdf

ches: Yeah at first I just made Vim spellcheck happy because I don't have an opinion, but I yielded after getting to the section on `sbt-datatype`. It's obviously in common use in industry vernacular so I'm happy to revert these to stay consistent.

src/reference/90-Developers-Guide/09-Launcher/01-Launcher-Getting-Started.md
```diff
@@ -116,11 +122,18 @@ interface class will be provided by the launcher, so it is only a
 compile-time dependency. If you are building with sbt, your dependency
 definition would be:
 
-Make the entry point to your class implement 'xsbti.AppMain'. An example
+```scala
+libraryDependencies += "org.scala-sbt" % "launcher-interface" % "1.0.0" % "provided"
+
+resolvers <+= sbtResolver
```
dwijnand: Could you make this `resolvers += sbtResolver.value`, it's the newer syntax.

src/reference/90-Developers-Guide/02-Coding-Guideline/00.md
eed3si9n: For the full explanation you should watch Josh's talk https://skillsmatter.com/skillscasts/3269-binary-resilience.
Both "causes" and "results in" suggestions are fine, I think but the larger point I am trying to convey is that `val` in trait reaches into the subclass code generation. This is undesirable for growable API because when we change the trait, it will require plugin authors to recompile their code against ours or else their plugin would blow up during runtime.

src/reference/90-Developers-Guide/02-Coding-Guideline/00.md
```diff
@@ -90,83 +90,87 @@ Use [MiMa][mima].
...
+
 - `val` or `var` in `trait` results to codegen at subclass and at the artificial `Foo\$class.\$init\$`.
 - `lazy val` results to codegen at subclass
 
 #### Abstract classes are also useful
 
 [To trait, or not to trait?][pins_trait127].
-It is less flexible compared to trait, but easier to maintain binary compatibility. It also has better Java interop.
+Abtract classes are less flexible compared to traits, but ease maintaining binary compatibility. They also have better Java interoperability.
```
eed3si9n: Are you using *ease* as verb here?

src/reference/90-Developers-Guide/02-Coding-Guideline/00.md
```diff
@@ -90,83 +90,87 @@ Use [MiMa][mima].
...
 
 #### Finalize the leaf classes
 
-If there's no need to keep the classes open, finalize it.
+If there's no need to keep a class open, finalize it.
 
 #### Typeclass and subclass inheritance
 
-Typeclass pattern with pure trait might be easier to maintain binary compatibility than subclassing.
+The typeclass pattern with pure traits might be ease maintaining binary compatibility more so than subclassing.
```
eed3si9n: *be ease*?

src/reference/90-Developers-Guide/02-Coding-Guideline/00.md
```diff
@@ -90,83 +90,87 @@ Use [MiMa][mima].
...
-### Other public API matter
+### Other public API matters
 
-Here are other guidelines about public API.
+Here are other guidelines about the sbt public API.
 
 #### Avoid Stringly-typed programming
 
-Define datatypes.
+Define data types.
```
eed3si9n: datatypes.

src/reference/90-Developers-Guide/02-Coding-Guideline/00.md
```diff
@@ -90,83 +90,87 @@ Use [MiMa][mima].
...
 
 #### Avoid overuse of `def apply`
 
 `def apply` should be reserved for factory methods
 in a companion object that returns type `T`.
 
-#### Use specific datatype (`Vector`, `List`, or `Array`) rather than `Seq`.
+#### Use specific data types (`Vector`, `List`, or `Array`), rather than `Seq`
```
eed3si9n: datatype

src/reference/90-Developers-Guide/02-Coding-Guideline/00.md
ches: Yes, is it hard to read?

dwijnand: Yeah: https://en.wiktionary.org/wiki/ease#Verb

src/reference/90-Developers-Guide/02-Coding-Guideline/00.md
ches: Thanks, "be" should have been removed. Fix incoming.

src/reference/90-Developers-Guide/02-Coding-Guideline/00.md
```diff
@@ -90,83 +90,87 @@ Use [MiMa][mima].
...
 
 #### Typeclass and subclass inheritance
 
-Typeclass pattern with pure trait might be easier to maintain binary compatibility than subclassing.
+The typeclass pattern with pure traits might be ease maintaining binary compatibility more so than subclassing.
 
 #### Avoid case classes, use sbt-datatype
 
-case class involves various codegen, which makes it harder to maintain binary compatibility over time.
+`case class` involves code generation that makes it harder to maintain binary compatibility over time.
```
eed3si9n: Y U no like *codegen* and *interop*? These are fairly common lingo for sbt developers, along with *bincompat*.

ches: Just my opinion: in formal public documentation it's best to expand abbreviations, at least on first use, and try to avoid jargon.

Keep in mind also that people reading this might not *yet* be sbt developers (e.g. me :smiley:).

src/reference/90-Developers-Guide/02-Coding-Guideline/00.md
eed3si9n: I don't think it conveys the same meaning as the original, since you can't just add abstract classes on top of a trait to ease the pain.

src/reference/90-Developers-Guide/02-Coding-Guideline/00.md
eed3si9n: I'm ok with expand-on-first-use rule, but I think jargons like these make the communication more effective since shorter words means less eyeball scrolling (or typing).

src/reference/90-Developers-Guide/02-Coding-Guideline/00.md
ches: > you can't just add abstract classes on top of a trait to ease the pain.

I don't feel my wording suggests that, or departs from the original in this regard, but let's try another—how about:

> Abstract classes are less flexible than traits, but traits pose more problems for binary compatibility. Abstract classes also have better Java interoperability.

src/reference/90-Developers-Guide/02-Coding-Guideline/00.md
```diff
@@ -90,83 +90,87 @@ Use [MiMa][mima].
...
 
-sbt-houserules comes with scalariform.
+sbt-houserules comes with scalariform for formatting source code consistently.
 
 #### Avoid procedure syntax
 
-Return `Unit`.
+Declare an explicit `Unit` return.
+
+#### Define typeclass instances in companion objects
```
eed3si9n: This is too strong compared to the original. Defining typeclass instance is not always possible or recommended.

src/reference/90-Developers-Guide/02-Coding-Guideline/00.md
ches: "First use" becomes slippery though in documentation as extensive as sbt's where readers may consume it in non-linear fashion, starting in different places depending on what they're looking for. Also when it's edited over time, by different people, the order tends to change.

Ultimately it's your call, so if you'd like specific instances reverted let me know.

src/reference/90-Developers-Guide/06-Datatype/00.md
```diff
@@ -5,12 +5,12 @@ out: Datatype.html
 Datatype
 --------
 
-sbt-datatype is a code generation library and an sbt autoplugin that generates
+`sbt-datatype` is a code generation library and an sbt autoplugin that generates
```
eed3si9n: Why sbt-datatype formatted this way? Shouldn't the monospace be reserved for code snippets and literals?

src/reference/90-Developers-Guide/02-Coding-Guideline/00.md
eed3si9n: Also monospace here doesn't make sense.

src/reference/90-Developers-Guide/02-Coding-Guideline/00.md
ches: How about "Use companion objects when defining typeclass instances"?

src/reference/90-Developers-Guide/06-Datatype/00.md
ches: Stylistic choice, my habit but I don't feel strongly either way.

One could raise similar bikeshedding over the use of "sbt" throughout the docs, that it's all-lowercase, whether it's ever capitalized at the beginning of a sentence (I say no, it's being treated like a brand mark), etc.

I don't particularly care either way, but what I would ask is that if you *do* feel strongly about it, please define a documentation style guide somewhere :smile:

src/reference/90-Developers-Guide/02-Coding-Guideline/00.md
eed3si9n: What do you think about adding a glossary section? I want to keep the developer section somewhat *raw*, similar to the way Dale and I might communicate on Gitter. Having a glossary might aid in potential sbt-dev jumping into our discussion on a code review etc.

ches: I think either way works, but if the monospace is removed then it ought to read:

> Case classes involves code generation that make it harder to maintain binary compatibility over time.

ches:
I pushed a few changes in response to feedback already, in commits that are ready for a `git rebase --interactive --autosquash`, so I'll squash them when everyone is satisfied. I skipped CI until then.

src/reference/90-Developers-Guide/06-Datatype/00.md
eed3si9n: I'm totally open to adding a style guide. *sbt* is always lower cased, and not monospaced when it refers to the project or the build tool in general. It is monospaced when it's referring to the shell script `sbt` that gets install on one's machine.

src/reference/90-Developers-Guide/02-Coding-Guideline/00.md
eed3si9n: Sorry, I meant Defining typeclass instance *in companion objects* is not always possible or recommended. So what I am trying to say is:

"It's ok to define typeclass instances in companion objects (as opposed to implicit converters, which is not ok)"

src/reference/90-Developers-Guide/02-Coding-Guideline/00.md
```diff
@@ -175,13 +179,13 @@ object FooID {
 }
 ```
 
-#### Implicit converter for syntax (enrich-my-library) should be imported
+#### Implicit conversions for syntax (enrich-my-library pattern) should be imported
```
eed3si9n: What's imported are the converters (functions from `A => B`), not conversions (applications to the functions), I think.

src/reference/90-Developers-Guide/02-Coding-Guideline/00.md
ches: Okay, becomes a little more wordy, but I think then the heading should be like:

> Define instances of typeclasses in their companion objects, when possible

Then, include an additional example of when it's okay/necessary *not* to do this.

src/reference/90-Developers-Guide/02-Coding-Guideline/00.md
eed3si9n: > Abstract classes are less flexible than traits, but traits pose more problems for binary compatibility. Abstract classes also have better Java interoperability.

+1

eed3si9n:
@ches Thanks for going through this.

src/reference/90-Developers-Guide/02-Coding-Guideline/00.md
eed3si9n: > Define instances of typeclasses in their companion objects, when possible

+1

src/reference/90-Developers-Guide/06-Datatype/00.md
ches: I feel the same way but my additional habit is monospacing when using something in the specific context of referring to its artifact, which I think is why it was instinctive here. Same with `launcher-interface` below.

src/reference/90-Developers-Guide/02-Coding-Guideline/00.md
ches: I think your distinction makes semantic sense, I just find the phrase "implicit conversion" used interchangeably in this manner across pretty much all Scala documentation I've read, and I've never seen "implicit converter" used… https://www.google.co.th/search?q=scala+implicit+converter

src/reference/90-Developers-Guide/06-Datatype/00.md
eed3si9n: Well, sort of yes and no. sbt-datatype in my head is a conceptual thing, like sbt is. It's not even a single artifact because it's made up of two parts: a library and a plugin: https://github.com/sbt/sbt-datatype/blob/master/build.sbt

src/reference/90-Developers-Guide/02-Coding-Guideline/00.md
ches: Putting it in quotes leads some Google results, but substantially less. There is some precedent in fairly "official" sources though (https://www.playframework.com/documentation/2.2.x/ScalaJson) so if you'd like to revert it I don't feel strongly.

eed3si9n: No, I stand corrected. [Programming in Scala 1st ed](https://www.artima.com/pins1ed/implicit-conversions-and-parameters.html) says *implicit conversion* and the SLS actually uses a wacky term called *view*:

> Implicit parameters and methods can also define implicit conversions called views. A *view* from type `S` to type `T` is defined by an implicit value which has function type `S=>T` or `(=>S)=>T` or by a method convertible to a value of that type.

This is because the *real* implicit conversions include other things like numeric widening, which I think Paul recently tweeted about: https://twitter.com/extempore2/status/703362270835834880

ches: :+1: Great reference find on the spec, and that Twitter dialogue is interesting too.

ches:
Let me try to come up with some agreeable wording by tomorrow on the remaining point about binary compat implications of traits, and then I'll squash these commits. :night_with_stars:

dwijnand:
Superseded (or superseeded as I normally say) by #318, which was merged.

PR #258, opened 2016-06-08
Use navigator.userAgent for more reliable os detection on Linux Firefox

sbrunk:
navigator.appVersion only returns "5.0 (X11)" on Firefox Ubuntu so people have to look for sbt Debian packages in the documentation. This navigator.userAgent seems to be more reliable on all platforms.


eed3si9n:
Thanks!

PR #262, opened 2016-07-13
sbt-hocon added

aalleexxeeii:
sbt-hocon (operations on HOCON data against reference configuration aggregated from project dependencies)


eed3si9n:
Thanks!

PR #266, opened 2016-08-10
fix specs2 url link. remove unused specs link

xuwei-k:
http://www.scala-sbt.org/0.13/docs/ja/Library-Dependencies.html

Library-Dependencies-ja


eed3si9n:
Thanks!

PR #282, opened 2016-10-03
Update comments indention of 02-Input-Tasks.md

keypointt:
comments indention should be the same as surrounding code


dwijnand:
Thanks!

PR #283, opened 2016-10-08
Use long key ID instead of short one

A1kmm:
You can verify the long key ID is the only one corresponding to the short key ID (at present) on the keyserver using the following command:

    gpg --dry-run --keyid-format long --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 642AC823

Using short key IDs is not safe, and there are already real-world attacks out there against this - see https://lkml.org/lkml/2016/8/15/445.


eed3si9n:
I've opted to use the full fingerprint, which is 2EE0EA64E40A89B84B2DF73499E82A75642AC823 - 59ea6b87967cbefd3885219307e6d5dafa1a77be

PR #284, opened 2016-10-13
Add sbt 0.13 syntax migration guide

dwijnand:


src/reference/01-General-Info/90-Changes/45-sbt-0.13-syntax-migration-guide.md
```diff
@@ -0,0 +1,105 @@
+---
+out: sbt-0.13-Syntax-Migration-Guide.html
```
eed3si9n: If we follow the convention from http://www.scala-sbt.org/0.13/docs/Migrating-from-sbt-07x.html this page should be `Migrating-from-sbt-012x`?

src/reference/01-General-Info/90-Changes/45-sbt-0.13-syntax-migration-guide.md
```diff
@@ -0,0 +1,105 @@
...
+---
+
+## sbt 0.13 Syntax Migration Guide
+
+### Introduction
+
+Before sbt 0.13 (sbt 0.9 to 0.12) it was very common to see in builds the usage of two aspects of sbt:
+
+* the "fishbone" operator: `<<=`, `<+=`, `<++=`
```
eed3si9n: I don't want to introduce "fishbone" term here ourselves especially now that I'm getting rid of it.
If we're going to name it, I would call them "key dependency operator".

src/reference/01-General-Info/90-Changes/45-sbt-0.13-syntax-migration-guide.md
```diff
@@ -0,0 +1,105 @@
...
+* the tuple enrichments (apply and map) for TaskKey's and SettingKey's (eg. `(foo, bar) map { (f, b) => ... }`)
+
+The release of sbt 0.13 (which was over 3 years ago!) introduced the `.value` macro which allowed for much
```
eed3si9n: s/macro/DSL/

src/reference/01-General-Info/90-Changes/45-sbt-0.13-syntax-migration-guide.md
```diff
@@ -0,0 +1,105 @@
...
+easier to read and write code, effectively making both aspects redudant and they were removed from the official
+documentation.
+
+As they will be removed in upcoming release of sbt 1.0.0 we've deprecated them in sbt 0.13.13, and here we'll
+help guide you to how to migrate your code.
+
+### Migrating simple expressions
+
+With simple expressions such as:
+
```
eed3si9n: Fenced code block should be used here (i.e. ```scala)

eed3si9n: Also let's not do the vertical alignment thing since that's not done anywhere else in this doc.

PR #285, opened 2016-10-20
Try and fix wrapping in syntax migration guide

dwijnand:


dwijnand:
Full screen on my laptop it looks fine. But not full screen it looks bad: 
image

PR #293, opened 2016-11-04
Github -> GitHub

SethTisue:
scratching that itch


eed3si9n:
My brain of course filtered it as "Github -> Github", and I was like HUH for a while?

SethTisue:
> A man after my own heart.

@dwijnand you might enjoy my very first Scala bug report ever: https://issues.scala-lang.org/browse/SI-725

PR #294, opened 2016-11-08
Use scala-seed as the example sbt new template

adamw:
As scala-seed will be the "official" minimal project template, I think it makes sense to use it instead of `eed3si9n/hello.g8`.

Also, `sbt new` with `scala-seed` doesn't ask for the Scala version, hence updating example output block.


eed3si9n:
Thanks!

PR #306, opened 2016-12-30
Update sbt Getting Started guide

eed3si9n:
This is a proposal to update the Getting Started guide a bit.
I've tried to remove obsolete information and things that mostly sbt core devs care about (over-emphasis on immutability and the return types of DSL expressions); and to update some terminology (sbt shell as opposed to "the interactive mode").

I've also added a new page that tries to explain WHY we need build.sbt not just *how* and *what*: https://github.com/sbt/website/blob/3b99844c8e975093c32f3c12417f4dbaaebe8b56/src/reference/00-Getting-Started/07A-Task-Graph.md

Let me know if I'm moving towards the right direction.

/cc @dwijnand, @pfn, @heathermiller, @tpolecat, @lihaoyi


src/nanoc/content/download.html
```diff
@@ -48,15 +48,9 @@
   </div>
 </div>
 
-<div clsas="arg_all">
+<div clsas="arc_all">
```
cb372: HTML typo here, although it was already there so maybe it's a feature :)

src/reference/00-Getting-Started/02-Hello.md
```diff
@@ -10,63 +10,64 @@ Hello, World
...
 
 ```scala
-lazy val root = (project in file(".")).
-  settings(
-    name := "hello",
-    version := "1.0",
-    scalaVersion := "$example_scala_version$"
+import Dependencies._
+
+lazy val root = (project in file("."))
```
cb372: Couldn't this build.sbt be simplified a bit, as it's the very first one we encounter? e.g.

* no need to define a project, just write all the settings at the top level
* no need to use inThisBuild

src/reference/00-Getting-Started/07A-Task-Graph.md
```diff
@@ -0,0 +1,294 @@
...
+
+**Note**: `scalacOptions` and `checksums` have nothing to do with each other, they
+are just two keys with the same value type, where one is a task.
+
+It is possible to compile a `build.sbt` that aliases `scalacOptions` to
+`checksums`, but not the other way. For example, this is allowed:
+
+```scala
+// The scalacOptions task may be defined in terms of the checksums setting
+scalacOptions := {
```
cb372: Worth pointing out that this can also be written more simply as `scalacOptions := checksums.value` ?

eed3si9n: Yes. I intentionally split the task dependencies from the task body, but I should add a section to explain that they are often combined in the wild.

src/reference/00-Getting-Started/07A-Task-Graph.md
```diff
@@ -0,0 +1,294 @@
...
+another task or setting. The value method is special and may only be
+called in the argument to `:=` (or, `+=` or `++=`, which we'll see later).
+
+As a first example, consider defining the `scalacOption` that depends on
+`update` and `clean` tasks. Here are the definitions of these keys (from [Keys](../sxr/sbt/Keys.scala.html)):
+
+```scala
+val scalacOptions = TaskKey[Seq[String]]("scalac-options", "Options for the Scala compiler.")
+val update = TaskKey[UpdateReport]("update", "Resolves and optionally retrieves dependencies, producing a report.")
+val clean = TaskKey[Unit]("clean", "Deletes files produced by the build, such as generated sources, compiled classes, and task caches.")
```
dwijnand: You should use the macro

src/reference/00-Getting-Started/09-Multi-Project.md
```diff
@@ -10,65 +10,57 @@ out: Multi-Project.html
...
 
-lazy val core = project
+lazy val core = (project in file("core"))
 ```
 
-The name of the val is used as the project's ID and base directory name.
-The ID is used to refer to the project at the command line. The base
-directory may be changed from the default using the in method. For
-example, the following is a more explicit way to write the previous
-example:
```
dwijnand: I think we should keep this.

src/reference/00-Getting-Started/07A-Task-Graph.md
```diff
@@ -0,0 +1,324 @@
...
+```
+
+This sets the name in terms of its previous value as well as the
+organization and version settings.
+
+### Inlining .value calls
+
+As explained above, `.value` is a special method that is used to express
+the dependency to other tasks and settings.
+To emphasize the fact that it is *not* a normal method call that are evaluated
```
dwijnand: You switch from singular "it" to plural "are" in this sentence.

eed3si9n: I can't agree verb numbers to save my life. I hope a55fc0d12cba1d58215a92884e72ea9204908d4c looks ok.

dwijnand: That's perfect.

src/reference/00-Getting-Started/09-Multi-Project.md
eed3si9n: Done 4c358f028b7b2503ca4b82920f2125b5eabb28ec

dwijnand: thanks

src/reference/00-Getting-Started/07A-Task-Graph.md
```diff
@@ -0,0 +1,324 @@
...
+  [Make]: https://en.wikipedia.org/wiki/Make_(software)
+  [Ant]: http://ant.apache.org/
+  [Rake]: https://ruby.github.io/rake/
+
+Task graph
+----------
+
+This page explains `build.sbt` definition in more detail.
+It assumes you've read [.sbt build definition][Basic-Def] and
+[scopes][Scopes].
```
evacchi: I like where you are going with this revised guide. I am wondering whether the *Scopes* section is a real requirement in order to understand the following. Would it be possible to ignore the Scope axis, then delve into those details later ?

src/reference/00-Getting-Started/07A-Task-Graph.md
```diff
@@ -0,0 +1,324 @@
...
+
+#### Benefits of hybrid flow-based programming
+
+Rather than thinking `settings` as a key-value map,
+a better analogy would be to think of it as a DAG (directed acyclic graph)
+of tasks where the edges denote *happens-before*. Let's call this the _task graph_.
+There are several motivation to organizing the build this way.
+
+First is de-duplication. With flow-based programming, a task is executed only once even when it is depended by multiple tasks.
+For example, even when multiple tasks along the task graph depend on `compile in Compile`,
```
evacchi: again, this is the only reference to a scope in the entire chapter. For the sake of the comparison with Make (which AFAIK does not have one such distinction, but I'm no expert) I'd avoid talking about scopes here, and introduce them later, as an addendum.

briantopping:
I posted the following on Gitter, but I see the conversation is happening here. I think it agrees with what @evacchi said above:
> One thought that comes to mind: What if `What's the point of the build.sbt DSL?` was moved closer to the first page of the docs, or even put on the first page? In other words, I think this section has a lot of value to set the context for “how to think” and things like scope build on this, not the other way around...

src/reference/00-Getting-Started/02-Hello.md
```diff
@@ -4,72 +4,77 @@ out: Hello.html
...
+      organization := "com.example",
+      scalaVersion := "$example_scala_version$",
+      version      := "0.1.0-SNAPSHOT"
+    )),
+    name := "Hello",
+    libraryDependencies += scalaTest % Test
   )
 ```
 
+**Note**: Don't worry if the above build file doesn't make sense to you.
```
philwills: I worry about this language a bit, might it be better to say something like:

Don't worry if the details of this build file aren't clear yet, we'll come back to them later.

src/reference/00-Getting-Started/05-Basic-Def.md
```diff
@@ -241,12 +200,11 @@ task key still creates a value of type `T` when the task executes.
 
 The `T` vs. `Task[T]` type difference has this implication: a setting can't
 depend on a task, because a setting is evaluated only once on project
-load and is not re-run. More on this in
-[more kinds of setting][More-About-Settings], coming up soon.
+load and is not re-run. More on this in [task graph][Task-Graph], coming up soon.
```
philwills: I'd be tempted to drop ",  coming up soon"

src/reference/00-Getting-Started/07A-Task-Graph.md
```diff
@@ -0,0 +1,324 @@
...
+end
+```
+
+The breakthrough made with Rake was that it used a programming language to
+describe the actions instead of the system commands.
+
+#### Benefits of hybrid flow-based programming
+
+Rather than thinking `settings` as a key-value map,
+a better analogy would be to think of it as a DAG (directed acyclic graph)
```
philwills: I like the convention of introducing an acronym, by using the full name, followed by the acronym in brackets, rather than vice versa. I think it's less off putting if you're not familiar with the acronym.

src/reference/00-Getting-Started/07A-Task-Graph.md
```diff
@@ -0,0 +1,324 @@
...
+
+```scala
+scalacOptions := {
+  val ur = update.value
+  val x = clean.value
+  ur.allConfigurations.take(3)
+}
+```
+
+**Note**: The values calculated above are nonsensical for `scalaOptions`,
```
philwills: This feels like a really weird thing to do. At least for me, on seeing an example like this, my instinct is to try and understand what it's trying to achieve, which would be pointless in this case. It would be a while before I noticed this note.

src/reference/00-Getting-Started/09-Multi-Project.md
```diff
@@ -10,65 +10,68 @@ out: Multi-Project.html
...
 
-lazy val core = project in file("core")
+lazy val core = project
 ```
 
-#### Common settings
+#### Build settings
 
 To factor out common settings across multiple projects,
 create a sequence named `commonSettings` and call `settings` method
```
philwills: This doesn't seem to line up with the title/example any more.  Also, I wonder if using a term like 'Build-wide settings'  might help to make the distinction clearer.

src/reference/00-Getting-Started/02-Hello.md
```diff
@@ -4,72 +4,77 @@ out: Hello.html
...
+
+lazy val root = (project in file("."))
+  .settings(
+    inThisBuild(List(
+      organization := "com.example",
+      scalaVersion := "$example_scala_version$",
+      version      := "0.1.0-SNAPSHOT"
+    )),
+    name := "Hello",
+    libraryDependencies += scalaTest % Test
```
lihaoyi: This is the first time I'm seeing this `% Test` syntax; is that the new standard way of doing it that I should be following?

Where does `scalaTest` come from? I've never seen that either, and feel it might be clearer if we generated the full `"org.scalatest" %% "scalatest" % "..."` since a reader would be able to guess that that's some kind of module coordinates, vs `scalaTest` being an opaque identifier.

src/reference/00-Getting-Started/02-Hello.md
```diff
@@ -4,72 +4,77 @@ out: Hello.html
...
   [Setup]: Setup.html
+  [Running]: Running.html
 
 Hello, World
 ------------
 
 This page assumes you've [installed sbt][Setup].
 
-### Create a project directory with source code
+### sbt new command
```
tpolecat: Would it make sense to simply say 

> This page assumes you've installed sbt 0.13.10 or later.

Since there are is no guidance given for creating a project by hand? This seems fine, although [`sbt new` does not work for me](https://github.com/foundweekends/giter8/issues/228) and I have a hard time believing I'm the only person with this problem. We'll find out I guess.

src/reference/00-Getting-Started/03-Directories.md
```diff
@@ -50,6 +46,10 @@ src/
 Other directories in `src/` will be ignored. Additionally, all hidden
 directories will be ignored.
 
+Source code can be placed in the project's base directory as
+`hello/app.scala`. However, most people don't do this for real projects;
+too much clutter.
```
lihaoyi: Perhaps we could phrase this as 
```diff
+Source code can be placed in the project's base directory as
+`hello/app.scala`, which may be useful for small one-file throw-away 
+projects, though for larger projects people tend to keep source files in the 
+relevant src/main/XXX folder to keep things neat
```

src/reference/00-Getting-Started/04-Running.md
```diff
@@ -53,21 +53,23 @@ quotes. For example,
 In this example, `testOnly` has arguments, `TestA` and `TestB`. The commands
 will be run in sequence (`clean`, `compile`, then `testOnly`).
 
+**Note**: Running in batch mode requires JVM spinup and JIT each time, so your build may run slower. For day-to-day coding, we recommend using the sbt shell.
```
lihaoyi: We could add "or to use the **Continuous build and test** feature below, which works from the external command-line and also keeps the JVM running to allows fast re-builds"

src/reference/00-Getting-Started/05-Basic-Def.md
```diff
@@ -2,104 +2,63 @@
...
-.sbt build definition
----------------------
+Build definition
+----------------
 
 This page describes sbt build definitions, including some "theory" and
 the syntax of `build.sbt`. It assumes you know how to [use sbt][Running]
 and have read the previous pages in the Getting Started Guide.
 
-### Three Flavors of Build Definition
```
lihaoyi: 🎆

src/reference/00-Getting-Started/02-Hello.md
```diff
@@ -4,72 +4,77 @@ out: Hello.html
...
+      version      := "0.1.0-SNAPSHOT"
+    )),
+    name := "Hello",
+    libraryDependencies += scalaTest % Test
   )
 ```
 
+**Note**: Don't worry if the above build file doesn't make sense to you.
 In [.sbt build definition][Basic-Def] you'll learn more about how to write
 a `build.sbt` file.
```
tpolecat: I think the **Setting the sbt version** section below might be removed or re-worded … the g8 template sets this for you, and I don't think there's any reason to mention that it's optional (I would prefer that it not be).

src/reference/00-Getting-Started/05-Basic-Def.md
```diff
@@ -2,104 +2,63 @@
...
-
-In addition, a build definition can contain files ending in `.scala`, located in the
-`project/` subdirectory of the base directory to define commonly used functions and values.
+This page discusses the `build.sbt` build definition.
 
 ### What is a Build Definition?
 
 After examining a set of directories and processing build definition files, sbt
-ends up with `Project` definitions.
+ends up with a set of `Project` definitions.
```
lihaoyi: I think this is slightly confusing because earlier, `project` was used to refer to the whole thing that was created by `sbt new`, and now we have multiple `Project`s inside the original `project`. Could we pick different words to distinguish these two concepts? Maybe call the inner ones sub-`Project`s or something? Consistently call the outer one a `build`? I don't know what would be best

This certainly confused me when I first started using SBT.

src/reference/00-Getting-Started/03-Directories.md
tpolecat: Is there any reason to mention this here? As you say, it's not a good practice.

src/reference/00-Getting-Started/03-Directories.md
```diff
@@ -61,7 +61,7 @@ form the complete build definition. See [organizing the build][Organizing-Build]
 ```
 build.sbt
 project/
-  Build.scala
+  Dependencies.scala
```
tpolecat: Above it says

> `project` can contain `.scala` files, which are combined with `.sbt` files

which I find absolutely baffling since `.sbt` files are not Scala files and `.scala` source files don't compose. Do we really need to mention this in the Getting Started guide?

src/reference/00-Getting-Started/05-Basic-Def.md
```diff
@@ -110,7 +69,7 @@ rather than complete Scala statements.
 `build.sbt` may also be
 interspersed with `val`s, `lazy val`s, and `def`s. Top-level `object`s and
 `class`es are not allowed in `build.sbt`. Those should go in the `project/`
-directory as full Scala source files.
+directory as Scala source files.
```
tpolecat: How are definitions in in the `.scala` made visible in the `.sbt` file? We need an example here.

src/reference/00-Getting-Started/05-Basic-Def.md
```diff
@@ -138,8 +97,8 @@ the value `"hello"`.
 If you use the wrong value type, the build definition will not compile:
```
tpolecat: Above the text says

> In this case, the returned `Setting[String]` is a transformation to add or replace the `name` key in sbt's map, giving it the value `"hello"`.

But we just said the map is immutable. Can we maybe not claim it's immutable?

src/reference/00-Getting-Started/05-Basic-Def.md
```diff
@@ -138,8 +97,8 @@ the value `"hello"`.
...
 
 ```scala
-lazy val root = (project in file(".")).
-  settings(
+lazy val root = (project in file("."))
+  .settings(
     name := 42  // will not compile
   )
 ```
```
tpolecat: Below it says

> `val`s and `def`s must be separated from settings by blank lines.

(a) is this true? if so there needs to be section explaining the blank line thing. And (b) we're not showing settings at the top level (they're all in `.settings(...)`) so this may be confusing.

src/reference/00-Getting-Started/05-Basic-Def.md
```diff
@@ -2,104 +2,63 @@
...
+    name := "Hello",
+    scalaVersion := "$example_scala_version$"
+  )
 ```
 
 Each project is associated with an immutable map (set of key-value pairs) describing the project.
 
 For example, one key is `name` and it maps to a string value, the name of
 your project.
-
```
lihaoyi: 🎆

src/reference/00-Getting-Started/05-Basic-Def.md
```diff
@@ -2,104 +2,63 @@
...
-Instead, the build definition creates a huge list of objects with type
-`Setting[T]` where `T` is the type of the value in the map. A `Setting`
-describes a *transformation to the map*, such as adding a new key-value
-pair or appending to an existing value. (In the spirit of functional
-programming with immutable data structures and values, a transformation
-returns a new map -- it does not update the old map in-place.)
-
-Here is how you associate the `Setting[String]` for the name of
-the project located in the current directory:
+The key-value pairs are listed under the `settings(...)` method as follows:
```
lihaoyi: Should we call this `.settings(...)` with a leading `.`? That's the nomenclature I tend to use when it's a method I need to call on something else, rather than a free-floating function already available in scope

src/reference/00-Getting-Started/05-Basic-Def.md
```diff
@@ -2,104 +2,63 @@
...
-lazy val root = (project in file(".")).
-  settings(commonSettings: _*).
-  settings(
-    name := "hello"
+lazy val root = (project in file("."))
+  .settings(
+    name         := "hello",
+    organization := "com.example",
+    scalaVersion := "$example_scala_version$",
+    version      := "0.1.0-SNAPSHOT"
```
lihaoyi: If we are going to teach people `inThisBuild` in the first example, we should continue with that in all further examples. If we think it's too complicated to explain, we should strip it out of the first example.

So far it seems to me able half-half with and without `inThisBuild`; my personal vote would be to strip it out, since I've written tons of SBT configs and haven't ever used it, but I don't mind as long as it's consistent throughout

src/reference/00-Getting-Started/07A-Task-Graph.md
```diff
@@ -0,0 +1,324 @@
...
+----------
+
+This page explains `build.sbt` definition in more detail.
+It assumes you've read [.sbt build definition][Basic-Def] and
+[scopes][Scopes].
+
+### What's the point of the build.sbt DSL?
+
+As we saw before, a [build definition][Basic-Def] consists of subprojects
+with a map called `settings` describing the subproject.
```
lihaoyi: "with some `settings` describing the subproject" would be clearer I think. The fact that it's a map is not very useful, especially since it's not a `scala.Map`, nor is it any "normal" map that someone may think of since it contains tons of dependency information and other things

src/reference/00-Getting-Started/07A-Task-Graph.md
```diff
@@ -0,0 +1,324 @@
...
+There's more to the story.
+
+What the `Setting` sequence encodes is tasks and the dependencies among them,
+similar to [Make][Make] (1976), [Ant][Ant] (2000), and [Rake][Rake] (2003).
+
+#### Intro to Make
```
lihaoyi: Should this be kicked down to the tail end of the file? "How to use" SBT by declaring dependencies and inspecting stuff should be first priority. Justifying our own design decisions, and a comparative analysis with other build tools, should come later. Perhaps even on a separate page.

As a newbies, I do not want a tour of 3 other build tools before figuring out how to use SBT

src/reference/00-Getting-Started/07A-Task-Graph.md
```diff
@@ -0,0 +1,324 @@
...
+
+Second is parallel processing. Using the task graph, the task engine can
+schedule mutually non-dependent tasks in parallel.
+
+Third is the separation of concern and the flexibility.
+The task graph lets the build user wire the tasks together in different ways,
+while sbt and plugins can provide various features such as compilation and
+library dependency management as functions that can be reused.
+
+### Declaring dependency to other tasks
```
lihaoyi: This should be the top of this file IMO

src/reference/00-Getting-Started/07A-Task-Graph.md
```diff
@@ -0,0 +1,324 @@
...
+  val o = organization.value
+  val v = version.value
+  s"project \$old from \$o version \$v"
+}
+```
+
+This sets the name in terms of its previous value as well as the
+organization and version settings.
+
+### Inlining .value calls
```
lihaoyi: I think these should come first, with the `{}` version coming after.

In particular, all the `{}` examples above are so trivial they should be inline. Artificially splitting them into two statements with no obvious necessity will just confuse people, who will think it's a convention/best-practice/requirement.

We should pick a better, more motivating example to make the `{}` example make sense. Maybe add a `println` in the middle of one to help debug/show stuff?

src/reference/00-Getting-Started/03-Directories.md
lihaoyi: ...or as @tpolecat we could also just not talk about it

src/reference/00-Getting-Started/01-Setup/a.md
```diff
@@ -9,28 +9,24 @@ out: Installing-sbt-on-Mac.html
...
-
-Download [ZIP][ZIP] or [TGZ][TGZ] package, and expand it.
-
-### Installing manually
+#### [Macports](http://macports.org/)
 
-See instruction to install manually.
+```
+\$ port install sbt
+```
```
dwijnand: As of October it does work on Windows (Cygwin): https://github.com/paulp/sbt-extras/pull/167

src/reference/00-Getting-Started/07A-Task-Graph.md
```diff
@@ -0,0 +1,324 @@
+---
+out: Task-Graph.html
+---
+
+  [Basic-Def]: Basic-Def.html
+  [Scopes]: Scopes.html
+  [Make]: https://en.wikipedia.org/wiki/Make_(software)
+  [Ant]: http://ant.apache.org/
+  [Rake]: https://ruby.github.io/rake/
+
```
lihaoyi: This section should start off defining the key terms:

- Task
- Setting

With one/two-liner descriptions. The details of setting dependencies, syntax, etc. can come later, but a paragraph that makes it clear "there are tasks and settings and they are similar but different things" would I think help greatly for anyone trying to understand subsequent exposition

dwijnand:
Thank you all for the feedback. It's really heartwarming to see people help out where possible.

src/reference/00-Getting-Started/07B-Appending-Values.md
```diff
@@ -0,0 +1,84 @@
...
+
+Assignment with `:=` is the simplest transformation, but keys have other
+methods as well. If the `T` in `SettingKey[T]` is a sequence, i.e. the key's
+value type is a sequence, you can append to the sequence rather than
+replacing it.
+
+- `+=` will append a single element to the sequence.
+- `++=` will concatenate another sequence.
+
+For example, the key `sourceDirectories in Compile` has a `Seq[File]` as its
```
lihaoyi: is `sourceDirectories in Compile` a key? 

Or is `sourceDirectories in Compile` a task and `sourceDirectories` the key? 

If `sourceDirectories in Compile` is a task, is the un-scoped `sourceDirectories` also a task?

This is not clear to me, even given this description, as someone who has used SBT for years. We need to be more precise and consistent with this terminology so we can drill into people the same set of standard terms

tpolecat:
The page on scopes is still really infuriating. Should I post comments here or open another issue?

src/reference/00-Getting-Started/07A-Task-Graph.md
tpolecat: It might be useful to demonstrate specifically that you can't prevent a task from being evaluated by guarding with an `if` statement. This behavior is totally jaw-dropping and it's important to understand.

src/reference/00-Getting-Started/02-Hello.md
eed3si9n: > I wonder, could we add some `rootProject` helper to simplify the project in file(".")? 

I think it's possible, but I prefer reducing the number of constructs so it's fewer things to learn, especially if it doesn't add much functionality.

> Also, does it really need to be `lazy`, or could we do without that?

`lazy` does matter because subprojects can refer to each other via `aggregate(...)`, `dependsOn(...)`, and via scoping. You may or may not need it, but "always put `lazy`" is easier than "well it depends.."

> It's also not clear to me what inThisBuild is... does that apply to every project and not just the root project? Can I add it to any subproject and have the same effect? 

`inThisBuild` puts the setting in the build scope, which applies to all subprojects in the build.
Yes, you can add `thisBuild`-scoped settings to any subproject.

> Would it make sense to put these in a separate `val commonSettings = ... `?

`commonSettings` is in fact the way that's [currently](http://www.scala-sbt.org/0.13/docs/Basic-Def.html#How+build.sbt+defines+settings) documented, but it's repeating the same settings scoped to each subproject. `thisBuild`-scoped has a few advantages:
1. Because the value is centralized to a setting, it makes it easier to programmatically change it. This is useful for `version` (e.g. sbt-release uses `version in ThisBuild`)
2. A newly created subproject will be correct by default without manually adding in the common settings.

src/reference/00-Getting-Started/07A-Task-Graph.md
eed3si9n: It's probably possible to push the scope page a few pages back, but at some point the reader would need to read it to grasp build.sbt fully.

eed3si9n:
@briantopping 

>  One thought that comes to mind: What if What's the point of the build.sbt DSL? was moved closer to the first page of the docs, or even put on the first page? In other words, I think this section has a lot of value to set the context for “how to think” and things like scope build on this, not the other way around...

I too think Task Graph would be helpful in setting the context, but I think it would be useful to describe sbt's shell environment usage first like `run` and `test`. The earliest I can move Task Graph would be before .sbt build definition page.

src/reference/00-Getting-Started/02-Hello.md
eed3si9n: Sounds good.

src/reference/00-Getting-Started/07A-Task-Graph.md
eed3si9n: I can move the Note above the example if it helps.

src/reference/00-Getting-Started/07A-Task-Graph.md
briantopping: I'm wondering if 'make' can be described as a singly-scoped SBT, and in turn, fully-scoped SBT as merely a "depth dimension" of makefiles? 

If things could be inexactly considered in those terms, the novice could put their mind at rest not to worry about scopes until later, but at least be mentally prepared for when they were presented.

briantopping:
@eed3si9n 

> I too think Task Graph would be helpful in setting the context, but I think it would be useful to describe sbt's shell environment usage first like run and test. The earliest I can move Task Graph would be before .sbt build definition page.

Would it be possible to create two different documentation "parts"? One of the things I find refreshing about the changes you have here it is almost devoid of forward references. I think that's hurt a lot of the existing docs. Forward references and tight sequencing allow docs to be concise, but beginners need to map from familiar paradigms. It can be *really* hard to do that if a single concept gets missed. 

Splitting the docs into an "inexact, easily mappable colloquial" part and a "concise cross-referenced (but still readable)" part is valuable because different people learn differently. People that learn easily from one style may be find the other style to be unusable, at least early on. Both need an entry to get to a basic conversational understanding of SBT, then they can both work from a single document (the concise reference). 

I also think this might open the door for additional documentation contributors. The current docs are tightly sequenced and there is a "wrong way" to present things, potentially excluding writers that may have a less exact presentation. This has it's own pitfalls, but I don't think it's worth dismissing for the reasons in the previous paragraph.

src/reference/00-Getting-Started/01-Setup/a.md
eed3si9n: ok. The launcher story is a mess on its own so I've split it up to https://github.com/sbt/sbt/issues/2889

src/reference/00-Getting-Started/02-Hello.md
eed3si9n: I think it makes sense to put the disclaimer. There's also a project that we're working on that turns select Giter8 template into a download, so we can try that too.

src/reference/00-Getting-Started/03-Directories.md
eed3si9n: I think we have to for the sake of completeness. This is a page titled Directories, and it talks about the places where sbt will look for `*.scala` file out of the box. Since sbt is coded to look for them the base directory, we should mention it.

Another reason to mention this is because it does become relevant/recommended for the meta-build. See http://www.scala-sbt.org/0.13/docs/Organizing-Build.html#sbt+is+recursive

src/reference/00-Getting-Started/03-Directories.md
eed3si9n: I can link to http://www.scala-sbt.org/0.13/docs/Organizing-Build.html

src/reference/00-Getting-Started/05-Basic-Def.md
eed3si9n: Also a link to http://www.scala-sbt.org/0.13/docs/Organizing-Build.html here as well.

src/reference/00-Getting-Started/05-Basic-Def.md
eed3si9n: Yes, I also find the immutability break dancing to be odd. I'll remove the mention of it.

To be fair, the original authors might be saying that even though the DSL exposes procedural interface (assigning to a variable-like thing), *internally* it's kept purely functional by translating the DSL to something like:

```scala
scala> Map.empty[Symbol, Int].updated('x, 1).updated('x, 2)
res0: scala.collection.immutable.Map[Symbol,Int] = Map('x -> 2)
```

I don't see how that can be relevant to sbt users.

src/reference/00-Getting-Started/05-Basic-Def.md
eed3si9n: No it's no longer true since 0.13.7, so we can replace it with a warning for ppl to upgrade to a recent version at the top of the page.

src/reference/00-Getting-Started/02-Hello.md
eed3si9n: I think the notation was there, but got documented in #15. I like it since it feels more typesafe.

`scalaTest` comes from `import Dependencies._`. The idea of putting module ids in a file is discussed here - http://www.scala-sbt.org/0.13/docs/Organizing-Build.html

src/reference/00-Getting-Started/05-Basic-Def.md
eed3si9n: sounds good.

src/reference/00-Getting-Started/05-Basic-Def.md
eed3si9n: `inThisBuild()` came in sbt 0.13.9, and hasn't been documented much.
I have to mull this over on whether to include this in all examples or not. Maybe we can wait till we discuss multi-projects.. /cc @dwijnand

src/reference/00-Getting-Started/07A-Task-Graph.md
eed3si9n: Yea. The narrative of "build definition" is

> Each project is associated with an immutable map (set of key-value pairs) describing the project.

so I was trying to build some continuity with it. I can go with "with some settings describing the subproject"

src/reference/00-Getting-Started/07A-Task-Graph.md
eed3si9n: I'm betting that people need to be informed about philosophical background of `build.sbt` DSL to write `build.sbt`.

I'm fine with pushing the section down to the bottom.

eed3si9n:
@tpolecat As per Scopes, if you have feedback on improving the *documentation* of scopes please open a new issue on sbt/website. I personally feel like the problem is with scoping itself, and we should discuss the solution for that on https://github.com/sbt/sbt/issues/1812.

src/reference/00-Getting-Started/07A-Task-Graph.md
eed3si9n: `{}` version is the cognitive training wheel to signal to the reader that `.value` is really not a normal Scala method call. There's a temporal line on the sand between `val x = ...` and `ur...` 

```scala
scalacOptions := {
  val ur = update.value    // update task happens-before scalacOptions
  val x = clean.value      // clean task happens-before scalacOptions
  // ---- begin scalacOptions ----
  ur.allConfigurations.take(3)
}
```

I can add the `println` demo.

eed3si9n: And `if` demo too.

src/reference/00-Getting-Started/07A-Task-Graph.md
evacchi: I agree with @briantopping, but I'm not sure how/where would you say that so that it satisfies the "Makefile-literate" without sounding too arcane to everybody else.

src/reference/00-Getting-Started/07B-Appending-Values.md
eed3si9n: When we discuss the build.sbt DSL, the nomenclature could be something like:

```
organization :=   { name.value }
-----------  --    -------------
key        operator    body
--------------------------------
  setting (or task) expression
```

But colloquially we often say "organization is a setting."

In this case, `sourceDirectories in Compile` is a scoped key of type `SettingKey[Seq[File]]`.

src/reference/00-Getting-Started/05-Basic-Def.md
dwijnand: I still believe one should use ThisBuild scope as much as possible, over the commonSettings approach (for one, the state builds faster as there's less to compute). Sadly, for implementation details, you often have to fallback to commonSettings for some things (eg. anything depending on baseDirectory, including transitively dependant).

As such, I would introduce inThisBuild, and fairly early, as I agree to a more complete example that can be stripped down, instead of a minimal one that has to be built up.

Though I would not put it nested in the .settings of the root project. Firstly, it's more nesting than necessary. Secondly it makes it look like there's a specific reason why it's there, which isn't true. (I'm my fantasy sbt, settings not nested in a project are automatically scoped in ThisBuild, but alas..)

src/reference/00-Getting-Started/07A-Task-Graph.md
dwijnand: I think the information is useful and interesting (or at least I find it that way). But I also think it should be pushed to down or later in the docs.

src/reference/00-Getting-Started/01-Setup/00.md
```diff
@@ -28,9 +27,8 @@ To create an sbt project, you'll need to take these steps:
 Ultimately, the installation of sbt boils down to a launcher JAR
 and a shell script, but depending on your platform, we provide
 several ways to make the process less tedious.  Head over to the
-installation steps for [Mac][Mac], [Windows][Windows],
-[Linux][Linux], or
-[manual installation][Manual-Installation].
+installation steps for [Mac][Mac], [Windows][Windows], or
+[Linux][Linux].
```
fommil: GNU/Linux

src/reference/00-Getting-Started/09-Multi-Project.md
```diff
@@ -10,65 +10,68 @@ out: Multi-Project.html
...
 on each project. Note `_*` is required to pass sequence into a vararg
 method.
 
 ```scala
-lazy val commonSettings = Seq(
-  organization := "com.example",
-  version := "0.1.0",
-  scalaVersion := "$example_scala_version$"
-)
+lazy val root = (project in file("."))
```
olafurpg: I suggest that the recommended default style for multi-project builds looks like this

```scala
organization in ThisBuild := "com.foo"
scalaVersion in ThisBuild := "2.12.1"
version      in ThisBuild := "0.1.0"
// Optionally, use inThisBuild(List(...)) instead of repeating `in ThisBuild`.
// I personally think `in ThisBuild` reads cleaner but I'm happy to use
// inThisBuild for the sake of following convention.

lazy val core = project
  .settings(
    // ...
  )
  .enablePlugins(
    // ..
  )

lazy val extension = project
  .settings(
    // ...
  )
  .dependsOn(
    core
  )
```
Most notably, there is no `root` project. Why?

- Projects are automatically aggregated, which keeps things DRY. It's annoying to create a new project and start hacking with `~compile/test` only to realize you forgot to aggregate your new project. I imagine this can be especially confusing for beginners. It's rare that I don't want to exclude a project from aggregation, and I would consider it to be an advanced use-case.
- The root project name automatically becomes the root directory name. The root project name is used in IDEs like IntelliJ. My `File > Open recent` list is currently clogged projects named `root`.

src/reference/00-Getting-Started/07A-Task-Graph.md
olafurpg: I agree that explaining how SBT works should be the first priority. sbt may be the first build tool the reader encounters, and the current layout might give the impression that you need experience with other build tools to understand sbt. That's not true at all! Comparison with other build tools could belong to a separate "Coming from Make/Rake/..." section.

src/reference/00-Getting-Started/05-Basic-Def.md
dwijnand: No. And that's my point. ("Which isn't true" above)

src/reference/00-Getting-Started/09-Multi-Project.md
dwijnand: I create root projects simply to give it a stable (versioned) I'd. Otherwise it becomes dependent (outside version control) on the directory name. I advise against calling it root for the same reason you do, see the sbt modules having roots called sbtRoot, ioRoot, etc.

Also, please note that inThisBuild additionally scopes the rhs dependant keys, not just the lhs key. So

inThisBuild(Seq(a := b.value))

Is

a in ThisBuild := (b in ThisBuild).value

src/reference/00-Getting-Started/05-Basic-Def.md
eed3si9n: Let's split up the discussion around `inThisBuild(...)` to https://github.com/sbt/website/issues/307

src/reference/00-Getting-Started/02-Hello.md
evacchi: > > I wonder, could we add some rootProject helper to simplify the project in file(".")?
> I think it's possible, but I prefer reducing the number of constructs so it's fewer things to learn, especially if it doesn't add much functionality.

using `file(".")` has always looked a bit odd to me; I understand what it means, but in other contexts "." as CWD is usually assumed when omitted. But I understand that SBT's usual default is to use the val identifier as the dir name (another random thought, a bit OT: is it really worth that bit of magic to save a few extra chars? what if did away with it and just use `Project` ?)

random suggestion: `val currentDir = file(".")` so you can write `lazy val root = project in currentDir` (and, possibly, even `lazy val subp = project in currentDir / "subp"`)

dwijnand: The other benefit if the project macro over Project(..) is that it ensures the in-code identifier for the project (the val name) matches the runtime id of the project. I.e. avoid

```scala
val fooUtils = Project("fooUtil", file("foo-utilities"))
```

src/reference/00-Getting-Started/09-Multi-Project.md
olafurpg: > I create root projects simply to give it a stable (versioned) id. Otherwise it becomes dependent (outside version control) on the directory name.

It's a shame that comes at the price of missing auto-aggregation, which I think many beginner users would appreciate. I think it's at least worth documenting that `root` projects are optional and leaving out the root build defines useful defaults for multi-project builds.

> Also, please note that inThisBuild additionally scopes the rhs dependant keys, not just the lhs key

Cool! I didn't know about that. Time to rewrite my builds then I guess :) 


I think it's safe to assume that >95% of directory names are

src/reference/00-Getting-Started/01-Setup/00.md
hedefalk: +1, don't upset Stallman.

src/reference/00-Getting-Started/02-Hello.md
lihaoyi: Yeah I'd prefer `% Test` over `% "test"`, but I just hadn't ever seen it before. If that's the way to do things now, then happy to standardize on it.

I feel like putting things in a separate `project/Dependencies.scala` file, so early in the "Getting Started" section, is a bit confusing. Remember, at this point we know *nothing* about SBT at all. I think dealing with `build.sbt` is quite enough to learn, and anyway seeing/familiarizing with the `"org.scalatest" %% "scalatest" % "..."` syntax is itself a valuable thing to do. 

I think we should only break things out into separate files when the build gets a bit bigger, messier and non-trivial, which could be documented but somewhere later than the first Getting Started section

src/reference/00-Getting-Started/07A-Task-Graph.md
lihaoyi: My feeling is that people can get by quite far without the philosophical background. Things like

- Assigning some settings to literals
- Adding a `scala-reflect` `libraryDependency` that depends on your `scalaVersion`
- Adding `unmanagedSourceDirectories`s based on your `scalaVersion`

can all be accomplished by cargo-culting, without needing much philosophy at all, and will probably get a new SBT user quite far.

src/reference/00-Getting-Started/07A-Task-Graph.md
lihaoyi: Yeah I'm not convinced that it's a good training wheel. I think *one* example with this artificial split - the one you give here, with all the comments - is sufficient to illustrate the point. The comments are what make it understandable at all: otherwise it's just confusing to see these trivial assignments that "appear" to do nothing

src/reference/00-Getting-Started/07A-Task-Graph.md
```diff
@@ -0,0 +1,324 @@
...
+are just two keys with the same value type, where one is a task.
+
+It is possible to compile a `build.sbt` that aliases `scalacOptions` to
+`checksums`, but not the other way. For example, this is allowed:
+
+```scala
+// The scalacOptions task may be defined in terms of the checksums setting
+scalacOptions := {
+  val cs = checksums.value
+  cs
```
lihaoyi: We need to have more realistic, motivated examples for these introductory snippets. In it's current state, it's somewhat too abstract, and reading it I'm not sure what i'm trying to learn. This is especially true because we never "run" the examples, so I can't even really "see" what the result of each of these things is.

Furthermore, if we come up with realistic examples for every code snippet, apart from teaching the newbies how the *graph DSL* works, it will simultaneously teach them how to perform basic tasks, so they'd learn more useful things for the same amount of work

src/reference/00-Getting-Started/07A-Task-Graph.md
```diff
@@ -0,0 +1,324 @@
...
+
+```scala
+// The checksums setting may not be defined in terms of the scalacOptions task
+checksums := {
+  val so = scalacOptions.value
+  so
+}
+```
+
+#### Defining a setting that depends on other settings
```
lihaoyi: What's the difference between the `Declaring dependency to other settings` above, and ` Defining a setting that depends on other settings` here? It sounds about the same to me, and they have similar examples.

src/reference/00-Getting-Started/07B-Appending-Values.md
```diff
@@ -0,0 +1,84 @@
...
+you do this.
+
+#### Tasks based on other keys' values
+
+You can compute values of some tasks or settings to define or append value for another task. It's done by using `Def.task` and `taskValue`, as argument to `:=`, `+=` or `++=`.
+
+As a first example, consider appending a source generator using the project base directory and compilation classpath.
+
+```scala
+sourceGenerators in Compile += Def.task {
```
lihaoyi: Despite writing many `sourceGenerators`, it's not clear what this whole `Def.task{}.taskValue` thingy is about, even after reading this snippet. Why can we not directly `+=` the result of `myGenerator`?

src/reference/00-Getting-Started/07B-Appending-Values.md
```diff
@@ -0,0 +1,84 @@
...
+  myGenerator(baseDirectory.value, (managedClasspath in Compile).value)
+}.taskValue
+```
+
+### Appending with dependencies: `+=` and `++=`
```
lihaoyi: This seems to have a simpler example than `Tasks based on other keys' values` above. IMO this should be the first example, with a bunch more examples, and the `Def.task` thing only later (and even then only if we think it really should be on this page and not later, and if so with a lot more exposition explaining what the purpose of the `Def.task{}.taskValue` dance is necessary)

src/reference/00-Getting-Started/07B-Appending-Values.md
```diff
@@ -0,0 +1,84 @@
...
+
+Other keys can be used when appending to an existing setting or task,
+just like they can for assigning with `:=`.
+
+For example, say you have a coverage report named after the project, and
+you want to add it to the files removed by clean:
+
+```scala
+cleanFiles += file("coverage-report-" + name.value + ".txt")
+```
```
lihaoyi: I'm not sure if it'll fit on this page, but short shell-transcripts showing the results of running some SBT command, given these configs, and showing the results would be very useful to someone trying to understand it. 

Even if we can't fit a full transcript, a descriptive summary 

> now, if you run `sbt clean`, it will now additionally delete any `coverage-report-myProjectName.txt` file that exists in the root folder of the project"

Would help round out this example. This applies to all the other examples above, aswell.

src/reference/00-Getting-Started/09-Multi-Project.md
```diff
@@ -10,65 +10,68 @@ out: Multi-Project.html
...
 
-The name of the val is used as the project's ID and base directory name.
-The ID is used to refer to the project at the command line. The base
-directory may be changed from the default using the in method. For
-example, the following is a more explicit way to write the previous
-example:
+The name of the val is used as the subproject's ID, which
+is used to refer to the subproject at the sbt shell.
+
+Optionally the base directory may be omitted if it's same as the name of the val.
```
lihaoyi: I've seen `lazy val fooBar = project.in(file("foo-bar"))`

But yes, agree that we should just stick with relying on the `project` macro to put it in a relevant folder. There could be one example showing how to rename it, but the default should be good enough for newbies

src/reference/00-Getting-Started/07B-Appending-Values.md
tpolecat: Agree, also baffled.

src/reference/00-Getting-Started/09-Multi-Project.md
tpolecat: To me the DRY behavior is shocking because Scala programs are usually equivalent up to α-conversion (i.e., the only thing that should matter about an identifier is that it is distinct from other identifiers). But here the name of the identifier determines the name of the directory relative to the project root, so if I refactor my build and rename `foo` then it changes the meaning of my build. So personally I think this needs to be documented in red flashing text and the DRY behavior shouldn't be mentioned at all. (And it should be removed from sbt along with all the macros, but I doubt I'll win those arguments.)

Also I think (hope) it's becoming more standard to put sub-projects under a `modules` directory just to keep things uncluttered. So `val foo = project.in(file("modules/foo"))` may become more common. Multi-project builds are the norm in my world.

src/reference/00-Getting-Started/02-Hello.md
tpolecat: Oh is that how that works? I was trying to figure out how to set the id. None of this stuff should be magic. The cognitive cost is much higher than the saved keystrokes.

dwijnand: It should be documented better.

Also with 2.11+ I hope to be able to make the macro more obvious (using quasiquotes rather than the more obtuse AST macro implementation).

tpolecat: `val foo = projectWithIdAndDirectoryTheSameNameAsTheIdentifierImAssignedTo`.

dwijnand: lol. Reminds me of Guava calling flatMap [transformAndConcat](https://google.github.io/guava/releases/20.0/api/docs/com/google/common/collect/FluentIterable.html#transformAndConcat-com.google.common.base.Function-)

dwijnand: Btw, in both current and in future sbt you're still free to avoid the project macro and use the more verbose and repetitive Project constructor/apply invocation:

```scala
val foo = Project("foo", new java.io.File("foo"))
```

tpolecat: I'm going to start using that, thanks!

dwijnand: Btw the settingKey, taskKey, and inputKey macros exist also for the same reason. So you might want to start using those too:

```scala
val foo = SettingKey[Int]("foo", "I do foo")
val bar = TaskKey[Int]("bar", "I do bar")
val baz = InputKey[Int]("baz", I do baz")
```

evacchi: I'd still advocate for making this explicit; I'd see no harm in typing "foo" a few times; in particular, (if it's not already) one could make `Project("foo") == Project("foo", new java.io.File("foo"))` so you'd just end up writing `val foo = Project("foo")`

more on this theme on this gist https://gist.github.com/evacchi/9173ec455a1dc5c460298a20977dad8c and on sbt-dev once  my post gets approved

dwijnand: I'm more of a DRY and conventions-over-configuration advocate.

src/reference/00-Getting-Started/02-Hello.md
evacchi: the only problem I see with `Test` (which I prefer over `"test"`) is that apparently you have to resort to strings as soon as you have multiple specifiers (e.g. `"it,test"` [see the manual](http://www.scala-sbt.org/0.13/docs/Testing.html))

src/reference/00-Getting-Started/02-Hello.md
evacchi: I understand that; if other people are interested we can discuss this further on `sbt-dev` :-)

src/reference/00-Getting-Started/01-Setup/a.md
djspiewak: @milessabin In my experience, sbt-extras is considerably more reliable at dealing with a number of cases than the vanilla launcher, including sbt version.  I've also seen occasional really bizarre dependency symptoms in someone else's build that were fixed by swapping from the vanilla launcher to sbt-extras.  It's still the standard, in my book.

src/reference/00-Getting-Started/02-Hello.md
djspiewak: It should be possible to add some implicits which allow multiple specifiers, even with this syntax.  For example: `scalaTest % (Test, Integration)`

evacchi: yes, that would work

src/reference/00-Getting-Started/09-Multi-Project.md
tpolecat: @djspiewak I find that a lot of cruft accumulates at the root of projects and you end up with stuff like [this](https://github.com/typelevel/cats) where it's not at all clear what's a sub-project and what's, uh, something else. I may misunderstand your point though. My point was just that I basically never want a project's identifier and base directory to be the same.

src/reference/00-Getting-Started/02-Hello.md
chrisphelps: Would it be useful to include the link to the basic def page there as well?

For people who are not approaching this as a tutorial, some sort of text like the following might be useful (though I'm not sure whether its best to have this here, where we have a basic build.sbt, or later in Basic-Def where we talk about the structure):

> If you are trying to make sense of an existing build.sbt which does not look like this, refer also to Bare-Def and Full-Def

In the past I've tried to bootstrap from existing examples, and where the tutorials differ its always unclear why and which way I should proceed.

src/reference/00-Getting-Started/06-Scopes.md
```diff
@@ -24,7 +24,7 @@ context, called a "scope."
 
 Some concrete examples:
 
-- if you have multiple projects in your build definition, a key can
+- if you have multiple projects (also called subprojects) in your build definition, a key can
   have a different value in each project.
 - the `compile` key may have a different value for your main sources and
   your test sources, if you want to compile them differently.
```
chrisphelps: A code example up here somewhere would be useful. I read through pretty much the whole page here wondering how I specified a scope. Some of the content in the "Referring to scopes in a build definition" section would be great to move up towards the top.

chrisphelps: Also making those examples more real, as @lihaoyi suggested elsewhere, would be way better than ones that are "pointless! just an example".

src/reference/00-Getting-Started/07A-Task-Graph.md
```diff
@@ -0,0 +1,324 @@
...
+
+First is de-duplication. With flow-based programming, a task is executed only once even when it is depended by multiple tasks.
+For example, even when multiple tasks along the task graph depend on `compile in Compile`,
+the compilation will be executed exactly once.
+
+Second is parallel processing. Using the task graph, the task engine can
+schedule mutually non-dependent tasks in parallel.
+
+Third is the separation of concern and the flexibility.
+The task graph lets the build user wire the tasks together in different ways,
```
chrisphelps: Does this mean "build author"?

src/reference/00-Getting-Started/07A-Task-Graph.md
chrisphelps: I like your "temporal line in the sand" example. You somewhat talk around this here and in the section in Custom-Settings where you show the execution order of the `.value` calls. Explicitly putting something like your temporal explanation here would be useful.

src/reference/00-Getting-Started/09-Multi-Project.md
```diff
@@ -106,9 +109,9 @@ you can control aggregation per-task. For example, to avoid aggregating
...
 
 ```scala
-lazy val root = (project in file(".")).
-  aggregate(util, core).
-  settings(
+lazy val root = (project in file("."))
+  .aggregate(util, core)
+  .settings(
     aggregate in update := false
   )
```
chrisphelps: Below there is an example

`lazy val core = project.dependsOn(util)`

As all the other examples here added `in file(".")`, should we be consistent in that example?

src/reference/00-Getting-Started/11-Custom-Settings.md
```diff
@@ -239,7 +239,7 @@ lazy val library = (project in file("library")).
       val old = sampleStringTask.value
       println("stopping...")
       Thread.sleep(500)
-      old
```
chrisphelps: There are two `sampleStringTask`s defined in this example, with the one referring to the other. Is that correct? Even if it is, would the example be clearer if we renamed one of those?

src/reference/00-Getting-Started/02-Hello.md
eed3si9n: Let's split the discussion on enhancing `% Test` DSL to this issue - https://github.com/sbt/sbt/issues/1441

src/reference/00-Getting-Started/03-Directories.md
lihaoyi: I'd vote for not talking about the `project/Foo.scala` format until later in the `Organizing-Build` section. After all, it's purely an organizational/neatness thing, and isn't necessary to learn so early

src/reference/00-Getting-Started/05-Basic-Def.md
```diff
@@ -2,156 +2,126 @@
...
+```
+sbt.version=$app_version$
+```
 
-In addition, a build definition can contain files ending in `.scala`, located in the
-`project/` subdirectory of the base directory to define commonly used functions and values.
+to use sbt $app_version$.
+Without this file sbt will chose the sbt version installed on the machine,
+but setting the sbt version in
+`project/build.properties` avoids any potential confusion.
```
tpolecat: The transition from the follow-along style of the previous chapter is slightly jarring here. Maybe something like this?

> ### Specifying the required sbt version
> As part of your build definition you will specify the version of sbt that your project requires. This allows people with different versions of the `sbt` launcher to build the same project with consistent results. To do this, your project should have a file named `project/build.properties` that specifies the required version as follows:
> ```
> sbt.version=$app_version$
> ```
> If the required version is not available locally, `sbt` launcher will download it for you.  If this file is not present, the `sbt` launcher will choose an arbitrary version, which is discouraged because it makes your project non-portable.

src/reference/00-Getting-Started/07A-Task-Graph.md
eed3si9n: good point. I added a couple of real life examples in 45aaa78 and 3f3a253.

src/reference/00-Getting-Started/07A-Task-Graph.md
eed3si9n: The ordering is now flipped in 45aaa78cc3c44dfe82272471f9cbc869240869c8.

eed3si9n:
I couldn't address all of the comments, but got to most of them on the new contents, so I'd like to merge this PR before the work week begins tomorrow.

Huge thanks to all of you who took time from the off time, read through the docs, and gave feedback! The result came out improved with all your suggestions and discussions.

If you have more ideas about either the documentation or the sbt itself, please post to https://groups.google.com/forum/#!forum/sbt-dev or send us a PR.

src/reference/00-Getting-Started/09-Multi-Project.md
```diff
@@ -10,65 +10,68 @@ out: Multi-Project.html
...
+      organization := "com.example",
+      scalaVersion := "$example_scala_version$",
+      version      := "0.1.0-SNAPSHOT"
+    )),
+    name := "Hello",
+    publish := (),
+    publishLocal := ()
+  )
 
 lazy val core = (project in file("core")).
```
eed3si9n: I've split this up to https://github.com/sbt/website/issues/308

tpolecat:
Great work on this @eed3si9n, thank you.

src/reference/00-Getting-Started/07B-Appending-Values.md
```diff
@@ -0,0 +1,84 @@
...
+not, sbt will complain. It might say *"Reference to undefined setting"*,
+for example. When this happens, be sure you're using the key in the
+[scope][Scopes] that defines it.
+
+It's possible to create cycles, which is an error; sbt will tell you if
+you do this.
+
+#### Tasks based on other keys' values
+
+You can compute values of some tasks or settings to define or append value for another task. It's done by using `Def.task` and `taskValue`, as argument to `:=`, `+=` or `++=`.
```
leifwickland: Either "as an argument" or "as agruments"

leifwickland: I think you need an article in "or append **a** value"

leifwickland: The comma before "as argument" is unneeded.

src/reference/00-Getting-Started/09-Multi-Project.md
```diff
@@ -10,73 +10,76 @@ out: Multi-Project.html
...
 
-The name of the val is used as the project's ID and base directory name.
-The ID is used to refer to the project at the command line. The base
-directory may be changed from the default using the in method. For
-example, the following is a more explicit way to write the previous
-example:
+The name of the val is used as the subproject's ID, which
+is used to refer to the subproject at the sbt shell.
+
+Optionally the base directory may be omitted if it's same as the name of the val.
```
leifwickland: Nitpick: Missing article: "if it's **the** same"

eed3si9n:
@leifwickland Thanks for the proofing/copyediting. Much appreciated!

src/reference/00-Getting-Started/05-Basic-Def.md
```diff
@@ -2,156 +2,129 @@
...
+we often call it a *subproject* in this guide.
 
-In `build.sbt` you might create a [Project](../api/sbt/Project.html) definition of
-the project located in the current directory like this:
+For instance, in `build.sbt` you define
+the subproject located in the current directory like this:
 
 ```scala
 lazy val root = (project in file("."))
+  .settings(
```
eed3si9n: There is a page called [Appendix: Bare .sbt build definition](http://www.scala-sbt.org/0.13/docs/Bare-Def.html), but I'm steering new users away from using that in build.sbt.

src/reference/00-Getting-Started/03-Directories.md
```diff
@@ -50,18 +46,31 @@ src/
 Other directories in `src/` will be ignored. Additionally, all hidden
 directories will be ignored.
 
+Source code can be placed in the project's base directory as
+`hello/app.scala`, which may be for small projects,
+though for normal projects people tend to keep the projects in
+the `src/main/` directory to keep things neat.
+The fact that you can place `*.scala` source code might seem like
```
martijnhoekstra: should be "The fact that you can place `*.scala` source code *in the projects the base directory*"?

src/reference/00-Getting-Started/07A-Scopes.md
```diff
@@ -354,3 +353,42 @@ has three axes). The entire expression
...
+      // Same as:
+      // organization in ThisBuild := "com.example"
+      organization := "com.example",
+      scalaVersion := "$example_scala_version$",
+      version      := "0.1.0-SNAPSHOT"
+    )),
+    name := "Hello",
+    publish := (),
+    publishLocal := ()
+  )
```
julienrf: This is confusing to me because `inThisBuild` is used *within* the `.settings` of a project. Is it possible to write this global definitions *outside* of a project definition?

Also, it would be super interesting to mention `inConfig` as well.

julienrf: BTW, maybe you should also explain the difference between `ThisBuild` and `Global`?

src/reference/00-Getting-Started/07A-Scopes.md
```diff
@@ -354,3 +353,42 @@ has three axes). The entire expression
...
+lazy val core = (project in file("core")).
+  settings(
+    // other settings
+  )
+
+lazy val util = (project in file("util")).
+  settings(
+    // other settings
+  )
+```
```
julienrf: Something that’s missing, IMHO: I would rather change the way we recommend scoping keys to several axes to this syntax: `foo in task in config in project` rather than `foo in (task, project, config)` because I never remember the right order of the axes in the latter version.

src/reference/00-Getting-Started/07B-Appending-Values.md
```diff
@@ -0,0 +1,84 @@
...
+#### When settings are undefined
+
+Whenever a setting uses `:=`, `+=`, or `++=` to create a dependency on itself
+or another key's value, the value it depends on must exist. If it does
+not, sbt will complain. It might say *"Reference to undefined setting"*,
+for example. When this happens, be sure you're using the key in the
+[scope][Scopes] that defines it.
+
+It's possible to create cycles, which is an error; sbt will tell you if
+you do this.
```
julienrf: I would move this section to the `Scopes` page.

src/reference/00-Getting-Started/06-Task-Graph.md
```diff
@@ -0,0 +1,427 @@
...
+
+### Summary
+
+The core data structure of the build definition is a DAG of tasks,
+where the edges denote happens-before relationships.
+`build.sbt` is a DSL designed to express dependency-oriented programming,
+or flow-based programming, similar to `Makefile` and `Rakefile`.
+
+The key motivation for the flow-based programming is de-duplication,
+parallel processing, and customizability.
```
julienrf: Maybe it would be interesting to also add a section explaining why it is possible to write:

~~~ scala
foo := bar.value
~~~

But not:

~~~ scala
val foo = bar.value
~~~

Though the following is possible:

~~~ scala
val foo = Def.setting(bar.value)
~~~

In other words: what’s the magic happening in `:=` that makes it possible to omit the surrounding `Def.setting`. Maybe that could help understanding the `Def.task(…).taskValue` trick, too…

PR #315, opened 2017-01-13
Document sLog

dwijnand:


src/reference/04-Howto/06-Howto-Logging.md
```diff
@@ -278,3 +278,16 @@ myTask := {
   log.warn("A warning.")
 }
 ```
+
+### Log messages in a setting
+
+Since settings cannot reference tasks, the special task `streams`
+cannot be used to provide logging during setting initialization.
+The recommented way is to use `sLog`:
```
eed3si9n: Could you open a separate issue/PR for that?

PR #343, opened 2017-04-03
Add sbt-liquibase-slick-codegen to community plugins list

daniel-shuy:
Added sbt-liquibase-slick-codegen (a plugin to generate Slick database schema code from a Liquibase changelog file)


daniel-shuy:
After adding plugins to 2 different categories (Test plugins and Database plugins), I noticed that there are 2 different styles/conventions:

1. plugin-name: <url> (description)
2. plugin-name (description): <url>

Should there be a standardized template, or am I just being OCD?

eed3si9n:
I think the majority is `plugin-name (description): <url>` but there are a number of style 1s. I agree that we should stick to one style, but I won't lose sleep over it. PR would be welcome tho.

PR #357, opened 2017-05-04
Document scope delegation in detail

eed3si9n:
_Scope delegation_ (`.value` lookup) is one of more trickier aspects of build.sbt DSL. This writeup builds on top of the existing Scopes page, and breaks down the notion of scope delegation into rules and a few realistic looking examples.


src/reference/00-Getting-Started/07A-Scopes.md
```diff
@@ -393,3 +375,178 @@ lazy val util = (project in file("util"))
...
+      (bar in Test).value + 1
+    },
+    bar in Compile := 1
+  )
+```
+
+Inside of `foo`'s setting body the dependency a scoped key `(bar in Test)` is declared.
+However, as you see, `bar in Test` is undefined in `projA`.
+sbt is able to still resolve `(bar in Test)` to another scoped key,
+and initialize `foo` as `2`.
```
DavidGregory084: I know it sounds obvious, but it would be a good idea to explain that this is because (if I understand correctly) the `Test` scope `extend`s the `Compile` scope. It might also be worth explaining what it means for a scope to extend another scope.

dwijnand:
https://github.com/eed3si9n/website/pull/1

src/reference/00-Getting-Started/07A-Scopes.md
```diff
@@ -393,3 +375,178 @@ lazy val util = (project in file("util"))
...
+Here is an example.
+
+```scala
+lazy val root = (project in file("."))
+  .settings(
+    inThisBuild(List(
+      organization := "com.example",
+      scalaVersion := "2.12.2",
+      version      := scalaVersion.value + "_0.1.0"
+    )),
```
evacchi: this seems to suggest that inThisBuild works/makes sense only in a `settings()` call, while (IIRC) it works anywhere

src/reference/00-Getting-Started/07A-Scopes.md
```diff
@@ -393,3 +375,178 @@ lazy val util = (project in file("util"))
     // other settings
   )
 ```
+
+#### Delegation in detail (.value lookup)
+
+Now that we've covered all the details of scoping, we can explain the `.value`
+lookup in detail. It's ok to skip this section at in the beginning.
```
nightscape: Remove `at`

src/reference/00-Getting-Started/07A-Scopes.md
```diff
@@ -393,3 +375,178 @@ lazy val util = (project in file("util"))
...
+- There's a special value `ThisBuild` scope (written as `{.}` in shell) only for Subprojects axis.
+- build.sbt can reference a scoped key using `.in(...)` method.
+
+Now let's suppose we have a build definition as follows:
+
+```scala
+lazy val foo = settingKey[Int]("")
+lazy val bar = settingKey[Int]("")
+
+lazy val projA = (project in file("a"))
```
nightscape: Is this `in` method the same as the one below?

src/reference/00-Getting-Started/07A-Scopes.md
dispalt: I just came here to write that

sjrd:
Very nice write-up! I knew most of this, but not as clearly as I do now.

It also made be realize I had misunderstood the difference between `ThisBuild` and `Global`. Trouble is: now I don't know anymore what's the difference between them! Could that be clearly explained, further than just "they're at different levels of precedence"?

dwijnand:
@sjrd Having a good reference to link to people to explain the difference between `ThisBuild` and `Global` has been on my list for a while.

Basically sbt doesn't just have multi-project support, it also has multi-build support. It's not a widely used feature. For example if you add a dependency on another project using `ProjectRef` (where you specify the path of some other project) that project belongs to another build that isn't `ThisBuild`.

As a rule of thumb I would suggest to **always** use `ThisBuild` over `Global`, using `Global` only when:

* defining the default value of a brand new key, or
* modifying an sbt engine generic key, like `cancelable in Global`

sjrd:
Huh! Wow. TIL.

Looks like this multi-build feature could be a huge can of worms. I kind of always assumed that `ProjectRef`s were causing sbt to always have a "separate world" for the other build. But now I understand they use the same world, and they share their `Global` scopes (but not their `ThisBuild` scopes). I'll probably have to revisit all my `in Global`s now ...

dwijnand:
Yeah, I migrated away from my `in Global` usage too.

Have a look at this little session playing with how `projects` and `project` work when you add another build:

```
> projects
[info] In file:/d/sbt-lm/
[info] 	   lm
[info] 	 * lmRoot
[info] In file:/d/sbt-io/
[info] 	   io
[info] 	   ioRoot
> lm/compile
.. snip
> io/compile
[error] Expected ID character
[error] Not a valid command: io
[error] Expected project ID
[error] Expected configuration
[error] Expected ':' (if selecting a configuration)
[error] Expected key
[error] Not a valid key: io
[error] io/compile
[error]   ^
> {file:/d/sbt-io}io/compile
[success] Total time: 0 s, completed 05-May-2017 11:04:51
> project {file:/d/sbt-io}
[info] Set current project to IO Root (in build file:/d/sbt-io/)
> io/compile
[success] Total time: 1 s, completed 05-May-2017 11:05:51
> project {file:/d/sbt-lm}
[info] Set current project to LM Root (in build file:/d/sbt-lm/)
> lm/compile
.. snip
```

metasim:
How about a variant of `inspect`, say `inspect precedence <key>`, that behaves like `inspect uses <key>`, but shows all shadowed resolutions?

jvican:
> Yeah, I migrated away from my in Global usage too.
  
Perhaps I can do a rewrite for this. I could detect if people use the project refs for external project, and if not rewrite `Global` into `ThisBuild`. Does this sound like a good idea?

jvican:
> How about a variant of inspect, say inspect precedence <key>, that behaves like inspect uses <key>, but shows all shadowed resolutions?
  
What do you mean by shadowed "resolutions"?

src/reference/00-Getting-Started/07A-Scopes.md
```diff
@@ -393,3 +375,178 @@ lazy val util = (project in file("util"))
...
+Next see the "Delegates". This is *all* of the possible delegate candidates
+listed in the order of precedence.
+
+- All the scopes with `projC` scoping on Subproject axis are listed first,
+  then `ThisBuild` (`{.}`), and `Global` (`*`).
+- Within a subproject, scopes with `Compile` scoping on Dependency configuration axis
+  are listed first, then falls back to `Global` (`*`) configuration.
+- Finally, Task axis scoping lists the given task value `console::` and the one without.
+
+Note that scope delegation feels similar to class inheritance in an object-oriented language,
```
jvican: I suggest two things:
  
1. Make this a top-level section.
2. Avoid the term "dynamic dispatch" and describe what it really means. The example given is great, but it's difficult to parse for people because it involves understanding what dynamic dispatch. I would assume lots of people don't know what it is.
3. Write a sbt example that shows this behaviour.

jvican: Forget about the sbt example. I missed that there's already one.

eed3si9n: I adopted your suggestion 1. I might work on 2 later.

eed3si9n:
Thanks everyone for all the suggestions and comments.
Special credit to @dwijnand for https://github.com/eed3si9n/website/pull/1, which I've mostly hand merged here - https://github.com/sbt/website/pull/357/commits/9675fdbb6b80beff2ab192f34085ba4d9ac70a63

PR #362, opened 2017-05-11
Added comment about running `run` or `package` for resources

reactormonk:
Should solve https://github.com/sbt/sbt/issues/1832

### DIFF
--- a/src/reference/04-Howto/03-Howto-Generating-Files.md
+++ b/src/reference/04-Howto/03-Howto-Generating-Files.md
@@ -103,10 +103,11 @@ resourceGenerators in Compile += Def.task {
 }.taskValue
 ```
 
-Executing `run` (or `package`) will add a file `demo` to `resourceManaged`,
-which is `target/scala-*/resource_managed"`. By default, generated resources are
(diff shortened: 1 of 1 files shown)


lightbend-cla-validator:
Hi @reactormonk,

src/reference/04-Howto/03-Howto-Generating-Files.md
```diff
@@ -103,6 +103,11 @@ resourceGenerators in Compile += Def.task {
 }.taskValue
 ```
 
+Executing `run` (or `package`) will add a file `demo` to `resourceManaged`,
+which is `target/scala-*/resource_managed"`. By default, generated resources are
+not included in the packaged source artifact. To do so, add them as you would
+other mappings. See [Adding files to a package][modify-package-contents].
```
jvican: Can we add a small line here saying: "Note that running `compile` is not sufficient for `resourceGenerators` to run."

reactormonk: I figured that would be a bit too much text, and this would be sufficient.

dwijnand:
Thanks, @reactormonk.

PR #364, opened 2017-05-14
Add missing reference to Appending-Values chapter

exlevan:
Link to scopes in [Appending-Values](http://www.scala-sbt.org/0.13/docs/Appending-Values.html#When+settings+are+undefined) chapter is rendered as `[scope][Scopes]`, this commit fixes that.


eed3si9n:
Thanks!

PR #368, opened 2017-06-08
Move crosscompile ++ doc

Timshel:
Hey,

I was searching for the `++` information and missed it while skimming the documentation and finally found it in a third party website. It's only when I started writting the PR to add it that I realized it was already present. 

I still made this pr because I belive I'm probably not the first one to miss it.


eed3si9n:
Thanks for the contribution!

PR #372, opened 2017-06-22
Remove experimental warning from testForkedParallel

jroper:
The `testForkedParallel` feature is almost 4 years old, and in the time that it has existed, there has never been a confirmed bug report against it. It shouldn't be considered experimental anymore.

### DIFF
--- a/src/reference/02-DetailTopics/02-Configuration/14-Testing.md
+++ b/src/reference/02-DetailTopics/02-Configuration/14-Testing.md
@@ -223,8 +223,8 @@ forked JVMs allowed to run at the same time by setting the limit on
 actions cannot be provided with the actual test class loader when a
 group is forked.
 
-In addition, forked tests can optionally be run in parallel.  This feature
-is still considered experimental, and may be enabled with the following setting :
+In addition, forked tests can optionally be run in parallel within the
+forked JVM(s), using the following setting:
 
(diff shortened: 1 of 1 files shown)


dwijnand:
I had a look in the OSS space using GitHub code search, and it doesn't look like many people actually use it:

* [17 instances of "testForkedParallel in Test := true"](https://github.com/search?l=Scala&q=%22testForkedParallel+in+Test+%3A%3D+true%22&type=Code&utf8=%E2%9C%93)
* [13 instances of "testForkedParallel := true"](https://github.com/search?l=Scala&q=%22testForkedParallel+%3A%3D+true%22&type=Code&utf8=%E2%9C%93)
* [5 instances of "testForkedParallel in ThisBuild := true"](https://github.com/search?l=Scala&q=%22testForkedParallel+in+ThisBuild+%3A%3D+true%22&type=Code&utf8=%E2%9C%93)

so I don't know how much we can actually claim it's bug-free, even 4 years in.

I don't feel strongly for or against dropping the warning.

jroper:
I just turned it on on a project, and it works with no problems. The problem with leaving the experimental warning there is that no one will use it.

eed3si9n:
Let's drop the warning since we have no plan of changing or removing the feature in sbt 1.

PR #382, opened 2017-07-18
add sbt-protobuf

xuwei-k:


dwijnand:
Thank you, Yoshida san.

PR #397, opened 2017-07-21
trim is too aggressive, right trim is better

JustinPihony:
The leading space is needed if you actually have arguments.

### DIFF
--- a/src/reference/04-Howto/20-Howto-Sequencing/04-Howto-Dynamic-Input-Task.md
+++ b/src/reference/04-Howto/20-Howto-Sequencing/04-Howto-Dynamic-Input-Task.md
@@ -58,7 +58,7 @@ lazy val root = (project in file("."))
   )
 ```
 
-\* Note that some tasks (ie. `testOnly`) will fail with trailing spaces, so a right trim (`.replaceAll("\\s+$", "")`) of the string built for `toTask` might be needed to handle empty `args`.\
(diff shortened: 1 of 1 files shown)


src/reference/04-Howto/20-Howto-Sequencing/04-Howto-Dynamic-Input-Task.md
```diff
@@ -58,7 +58,7 @@ lazy val root = (project in file("."))
   )
 ```
 
-\* Note that some tasks (ie. `testOnly`) will fail with trailing spaces, so a `trim` of the string built for `toTask` might be needed to handle empty `args`.
+\* Note that some tasks (ie. `testOnly`) will fail with trailing spaces, so a right trim (`.replaceAll("\\s+$", "")`) of the string built for `toTask` might be needed to handle empty `args`.\
```
eed3si9n: $ sign needs to be escaped to `\$` in Pamflet.

eed3si9n (changes requested):

src/reference/04-Howto/20-Howto-Sequencing/04-Howto-Dynamic-Input-Task.md
JustinPihony: Fixed

eed3si9n (approved):

Issue #180, opened 2015-07-31
Search Engine Optimization


eed3si9n:
Currently 0.12.4 documentation shows up frequently.
- https://support.google.com/webmasters

Maybe build sitemap.xml?

eed3si9n:
I wonder if it helps to rename old sitemaps temporarily until we have the new one.

eed3si9n:
https://www.sitemaps.org/protocol.html

```xml

<?xml version="1.0" encoding="UTF-8"?>

<sitemapindex xmlns="http://www.sitemaps.org/schemas/sitemap/0.9">

   <sitemap>

      <loc>http://www.example.com/sitemap1.xml.gz</loc>

      <lastmod>2004-10-01T18:23:17+00:00</lastmod>

   </sitemap>

   <sitemap>

      <loc>http://www.example.com/sitemap2.xml.gz</loc>

      <lastmod>2005-01-01</lastmod>

   </sitemap>

</sitemapindex>
```

eed3si9n:
We haven't rolled out sitemap to robot.txt yet. As of July 27, 2017, the Google search for "sbt override dependencies" looks like this:

![sbt_override_dependencies_-_google_search](https://user-images.githubusercontent.com/184683/28677800-9bf64934-72bc-11e7-86fc-a0f7e545ba26.png)

PR #410, opened 2017-07-27
Resurrect site map

eed3si9n:
Fixes sbt/website#180


eed3si9n:
Last modified date won't work for Travis since git doesn't store dates.
I need to hard code something per version for old files.

PR #431, opened 2017-08-09
How to take an action on startup

eed3si9n:


eed3si9n:
http://www.scala-sbt.org/1.0/docs/Howto-Startup.html

PR #440, opened 2017-08-12
use releaseStepCommand instead of releaseStep and Command.process

xuwei-k:
https://github.com/sbt/sbt-release/blob/v1.0.6/src/main/scala/ReleasePlugin.scala#L92

`sbt.Command.process` changed and removed

- https://github.com/sbt/sbt/blob/v1.0.0/main-command/src/main/scala/sbt/Command.scala
- https://github.com/sbt/sbt/commit/1bf50e10c835e53e6ced74e131353d2fc5871ac0#diff-7e3f43c68fee15d3791010ec9c65b951R92
- https://github.com/sbt/sbt/commit/e83564a6b71639c7e51bbc002c2d083cf8c9ab5a#diff-7e3f43c68fee15d3791010ec9c65b951L91

### DIFF
--- a/src/reference/01-General-Info/05-Using-Sonatype.md
+++ b/src/reference/01-General-Info/05-Using-Sonatype.md
@@ -258,10 +258,10 @@ To automate the publishing approach above with the [sbt-release plugin]
 
 ```
 ...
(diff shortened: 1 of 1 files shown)


eed3si9n (approved):
Thanks!

PR #468, opened 2017-09-01
SbtExclusionRule doesn't exist anymore

manub:
This update the documentation to use `ExclusionRule` instead inside `excludeDependencies`.

I removed the reference to the SBT version in which this class started as IMHO doesn't add much in 1.0.x docs but I'll be happy to put it in place again if needed.


dwijnand (approved):
LGTM, thanks.

PR #474, opened 2017-09-05
use "onFailure" instead of "-"

xuwei-k:
- https://github.com/sbt/sbt/blob/v1.0.1/main-command/src/main/scala/sbt/BasicCommandStrings.scala#L179
- https://github.com/sbt/sbt/commit/86ae3c8c59df546dadb8c62901bac50c175af78a


dwijnand (approved):
TIL about onFailure! LGTM

PR #491, opened 2017-09-22
use Lightbend favicon and blue/orange color scheme

eed3si9n:
blue-orange


dwijnand:
no wrench?

eed3si9n:
@dwijnand now that we have _the_ wrench, I don't think we need any more.

dwijnand:
what's _the_ wrench?

eed3si9n:
The top-left logo will stay, but it doesn't render well as small size. See 

before

so I am using the Lightbend favicon:

favicon_screenshot

dwijnand:
I don't think that's too bad. I think it would benefit if the background were transparent rather than white, like GitHub's favicon:
image

PR #498, opened 2017-09-27
Recover Macro Projects shell snippets

dwijnand:
These were lost in the reStructuredText to Markdown conversion in
57bb93275be27eb6c2a58348204bacf1a0a5f90d.

Also update the resulting output.

### DIFF
--- a/src/reference/02-DetailTopics/02-Configuration/09-Macro-Projects.md
+++ b/src/reference/02-DetailTopics/02-Configuration/09-Macro-Projects.md
@@ -91,6 +91,12 @@ object Usage {
 
 This can be then be run at the console:
 
+```text
+$ sbt
+> macroSub/test:run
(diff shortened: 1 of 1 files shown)


src/reference/02-DetailTopics/02-Configuration/09-Macro-Projects.md
```diff
@@ -40,7 +40,7 @@ lazy val core = (project in file("core"))
     // other settings here
   )
 
-lazy val macroSub = (project in file("macro"))
+lazy val macroSub = Project("macro", file("macro"))
```
eed3si9n: I'd rather not use old ctor here. Could we update the example to use `macroSub` instead?

eed3si9n (changes requested):
comment above.

PR #502, opened 2017-09-28
What does sbt stand for, and how should I spell it?

eed3si9n:


dwijnand (changes requested):
See my tweaks in https://github.com/eed3si9n/website/pull/2.

dwijnand (approved):

PR #512, opened 2017-10-05
[1.x-beta] server documentation

eed3si9n:
Here's the start of server documentation, targeting the 1.x-beta branch, which will becomes immediately available on http://www.scala-sbt.org/1.x-beta/docs/.

The idea is to keep my topic branch going, and re-send the PR to 1.x once sbt 1.1.0 ships.


dwijnand (commented):
https://github.com/eed3si9n/website/pull/3/files

PR #515, opened 2017-10-06
[1.x-beta] Scope path notation docs

eed3si9n:
Here's the start of scope path notation docs, targeting the 1.x-beta branch, which will becomes immediately available on http://www.scala-sbt.org/1.x-beta/docs/.

Not all `in` are updated, but migration guide is added and also the scoping docs in the Getting Started guide is updated.


dwijnand (changes requested):
I don't think it should be called "scope path notation" for two reasons. 1. "path" indicates too strongly that it's hierarchical, when it's not, it's difference axes (of evil). 2. "Slash syntax" is more intuitive; it's what you would say causally when chatting to people about recent features of sbt. "Oh yeah we've improved the situation by adding unifying the build definition and shell syntaxes. Instead of using 'in' in code and slashes, colons and double colons in the shell it's now just slashes for both. For example: `Test / compile`. It's the new slash syntax."

dwijnand (changes requested):
https://github.com/eed3si9n/website/pull/4

dwijnand (approved):

PR #523, opened 2017-10-22
Fixed path for adding a plugin

laughedelic:
Was is a typo or I just didn't get what is meant? 
Noticed after this [SO question](https://stackoverflow.com/q/46860492/736957).


eed3si9n:
Thanks!

PR #530, opened 2017-10-26
Create API symlink for version 1.x, create "release" symlink pointing to "1.x"

dec5e:
Fixing API symlinks in proper place instead of sbt/sbt.github.com#42.


eed3si9n (approved):

dwijnand:
Thanks @dec5e

PR #537, opened 2017-10-29
use streams.value.cacheDirectory

xuwei-k:
`sbt.Keys.cacheDirectory` removed

https://github.com/sbt/sbt/commit/a0193d28ea1bf25824bc41b1e08ba128cd9041f3#diff-46751185767542b1d052bc42435916bdL152


dwijnand:
Thanks for all the PRs @xuwei-k!

Issue #527, opened 2017-10-25
0.13 docs don't link to 1.0.x


raboof:
When on http://www.scala-sbt.org/0.13/docs/index.html it seems like you're on the 'Latest version', even though you're not.

Philippus:
there's an option missing on:
https://github.com/sbt/sbt.github.com/blob/master/1.x/docs/index.html#L107
and
https://github.com/sbt/sbt.github.com/blob/master/0.13/docs/index.html#L105

the option is there on:
https://github.com/sbt/sbt.github.com/blob/master/0.12.4/docs/index.html#L83

`set-versions.js` depends on the option to do its magic

eed3si9n:
Related I sent in https://github.com/sbt/website/pull/507 to display outdated warning, but closed it because we thought it's still too soon.

Philippus:
closed by sbt/website#541 and related PR's

PR #542, opened 2017-11-01
Take selected version from url, closes sbt/website#527

Philippus:


eed3si9n:
_sphinx seems like really old stuff. Probably not being used.

Philippus:
There's only nanoc and sphinx in the branch. so at least one of them should be used I guess. :)

PR #541, opened 2017-11-01
Take selected version from url

Philippus:

PR #543, opened 2017-11-02
let set-versions.js and versions.js sync

eed3si9n:
@Philippus updated `set-versions.js` in https://github.com/sbt/website/pull/541, but that didn't seem to deploy: https://github.com/sbt/sbt.github.com/blob/7cbb54a55a2ab953a555d2c5d29903fa990ec810/assets/set-versions.js

This attempts to fix the syncing.


eed3si9n:
Sync worked - https://github.com/sbt/sbt.github.com/commit/61a6464ea4cc55f4cec6d6181624eaa78eeb4cea

PR #544, opened 2017-11-02
Restore var select

eed3si9n:
The version selector regressed in https://github.com/sbt/website/pull/541.

### DIFF
--- a/src/nanoc/content/set-versions.js
+++ b/src/nanoc/content/set-versions.js
@@ -9,13 +9,16 @@ $(document).ready(function() {
    // the selected version is in the pathname of the current url
    var selected = document.location.pathname.split('/')[1];
 
+   // get the version drop-down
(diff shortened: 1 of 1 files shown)


Philippus:
sorry about that! I almost had the same PR ready. :)

src/nanoc/content/set-versions.js
```diff
@@ -9,13 +9,18 @@ $(document).ready(function() {
    // the selected version is in the pathname of the current url
    var selected = document.location.pathname.split('/')[1];
 
+   // get the version drop-down
+   var select = $("#versions");
+   // clear the options, which should only include the current version
+   select.html('');
```
Philippus: is the clearing needed?

eed3si9n:
Merging without Travis since it's js file update is now validated.

Issue #555, opened 2017-11-19
Built-in keys link for 1.x build definition docs broken


NabilZaman:
The following [link](http://www.scala-sbt.org/1.x/sxr/sbt/Keys.scala.html)
on the build definition [page](http://www.scala-sbt.org/1.x/docs/Basic-Def.html)
is broken in the [Built-in keys](http://www.scala-sbt.org/1.x/docs/Basic-Def.html#Built-in+Keys) section

NabilZaman:
Also broken in [Library-Dependencies](http://www.scala-sbt.org/1.x/docs/Library-Dependencies.html)

PR #565, opened 2017-12-16
Fix broken links of build-in keys in docs (#555)

jiminhsieh:
Fix all of broken links of build-in keys in docs.


dwijnand (approved):
Thanks!

PR #567, opened 2017-12-19
Add a "Support" tab

dwijnand:
Inspired by https://www.lagomframework.com/support.html

### DIFF
--- a/src/nanoc/content/support.md
+++ b/src/nanoc/content/support.md
@@ -1,0 +1,29 @@
+---
+title: Support
+---
+
+[Lightbend]: https://www.lightbend.com
+
+## Support
+
+[Lightbend][] sponsors sbt and encourages contributions from the active community. Enterprises can adopt it for
+mission critical systems with confidence because Lightbend stands behind sbt with commercial support and
+services.
+
+[ask]: https://stackoverflow.com/questions/ask?tags=sbt
+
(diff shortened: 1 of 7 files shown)


src/nanoc/content/support.md
```diff
@@ -0,0 +1,29 @@
+---
+title: Support
+---
+
+[Lightbend]: https://www.lightbend.com
+
+## Support
```
eed3si9n, line 7: Could you add link to http://www.scala-sbt.org/community.html#how-can-I-get-help and deduplicate it?

eed3si9n (changes requested):

src/nanoc/content/support.md
dwijnand: I'm not a fan of the duplication too, but it (generally) looks like what's been done between https://www.lagomframework.com/support.html and https://www.lagomframework.com/get-involved.html.

What kind of deduplification did you mean? Because if you mean point this page to that, then it would be quite a bare page.

eed3si9n: Sorry, I meant to say adding link _from_ community.html, not _to_. 
Get involved should talk mostly about ppl helping sbt, and the new Support can talk about sbt helping ppl.

dwijnand: Oh right. Sure sounds good. Will do.

eed3si9n (approved):

PR #581, opened 2018-01-22
Emphasize that `%%` uses the binary Scala version

avdv:


dwijnand (approved):
Thanks!

xuwei-k:
This changes are not enough https://github.com/sbt/website/issues/456

dwijnand:
Sorry, Yoshida-san, why is this not enough?

If the concern is the olden days (2.9.x? early 2.10.x?) where the standard library wasn't binary compatible, then I think we needn't worry about that. IMO

eed3si9n:
The suggested change in this PR is fine.

As per  #456, let's talk about it there.

PR #611, opened 2018-02-05
added: how to use proxy repositories with credentials

lenalebt:
fixes https://github.com/sbt/sbt/issues/3844


eed3si9n (approved):
Thanks!

PR #613, opened 2018-02-08
Replace `Global` scope components with `Zero`

dkim:
Since sbt 1.0, `Global` as a scope component has been renamed `Zero` so as not to get confused with `Global` as a scope whose axes are all set to `Zero`.


eed3si9n (approved):

eed3si9n:
Thanks!

PR #617, opened 2018-02-10
fix pollInterval example

xuwei-k:
- https://github.com/sbt/sbt/blob/v1.1.1/main/src/main/scala/sbt/Defaults.scala#L251
- https://github.com/sbt/sbt/blob/v1.1.1/main/src/main/scala/sbt/Keys.scala#L143
- https://github.com/sbt/sbt/commit/0daca42c29188a511fd85bf4826dc05dac258b4b


dwijnand:
Thanks as always Yoshida-san.

PR #625, opened 2018-02-22
tweak wording on landing page

SethTisue:
"the shell" reads as a reference to the Unix shell. I think this
wording is a little clearer

### DIFF
--- a/src/nanoc/layouts/landing.html
+++ b/src/nanoc/layouts/landing.html
@@ -48,7 +48,7 @@
               <div id="main" class="text-center">
                 <h1>The interactive build tool</h1>
                 <span class="statement">
-                Use Scala to define your tasks. Then run them in parallel from the shell.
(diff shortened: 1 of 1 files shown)


src/nanoc/layouts/landing.html
```diff
@@ -48,7 +48,7 @@
               <div id="main" class="text-center">
                 <h1>The interactive build tool</h1>
                 <span class="statement">
-                Use Scala to define your tasks. Then run them in parallel from the shell.
+                Define your tasks in Scala. Run them in parallel from an interactive shell.
```
dwijnand: how about "from sbt's interactive shell"?

SethTisue: amended

eed3si9n (approved):

dwijnand (approved):
thanks Seth

SethTisue:
if you ever need a little extra space you can omit “your”

PR #628, opened 2018-02-26
Remove what looks like a typo in "Using Sonatype"

dwijnand:


SethTisue:
"I have discovered a truly marvelous example of this, which these parentheses are too narrow to contain."

PR #642, opened 2018-03-15
explain named pipe mode

martijnhoekstra:
This is a bit verbose, but the best I could come up with. Also, I'm not 100% sure it's correct - it describes my current understanding of the rather complex situation.

### DIFF
--- a/src/reference/02-DetailTopics/01-Using-sbt/07-sbt-server.md
+++ b/src/reference/02-DetailTopics/01-Using-sbt/07-sbt-server.md
@@ -90,11 +90,35 @@ Content-Length: ...\r\n
 }
 ```
 
+### Server modes
+
+Sbt server can run in two modes, which differ in wire protocol and initialization. The default mode since sbt 1.1.x is *domain socket mode*, which uses either Unix domain sockets (on Unix) or named pipes (on Windows) for data transfer between server and client. In addition, there is a *TCP mode*, which uses TCP for data transfer.
+
+The mode which sbt server starts in is goverened by the key `serverConnectionType`, which can be set to `ConnectionType.Local` for domain socket/named pipe mode, or to `ConnectionType.Tcp` for TCP mode.
+
 ### Server discovery and authentication
 
-To discover a running server and to prevent unauthorized access to the sbt server, we use a *port file* and a *token file*.
+To discover a running server, we use a *port file*.
 
-By default, sbt server will be running when a sbt shell session is active. When the server is up, it will create a file called the *port file*. The port file is located at `./project/target/active.json` relative to a build and contains something like:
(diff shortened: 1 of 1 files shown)


dwijnand:
nice, @martijnhoekstra! thank you very much for spending the time to rework this part of the docs for the next person. I'm going to request @eed3si9n reviews this, seeing as it's his baby :)

src/reference/02-DetailTopics/01-Using-sbt/07-sbt-server.md
```diff
@@ -104,13 +104,19 @@ By default, sbt server will be running when a sbt shell session is active. When
...
+or 
 
-1. That the server is (likely) running.
-2. That the server is running on port 5010.
-3. The location of the token file.
+```json
+{
+  "uri":"local:sbt-server-0845deda85cb41abdb9f",
+}
+```
```
eed3si9n: Assuming more people are on Mac and Linux, it might be better to use domain socket as the first example, and move TCP to later.

```json
{"uri":"local:///Users/someone/.sbt/1.0/server/0845deda85cb41abdb9f/sock"}
```

src/reference/02-DetailTopics/01-Using-sbt/07-sbt-server.md
```diff
@@ -126,7 +132,7 @@ The `uri` field is the same, and the `token` field contains a 128-bits non-negat
 
 To initiate communication with sbt server, the client (such as a tool like VS Code) must first send an [`initialize` request][lsp_initialize]. This means that the client must send a request with method set to "initialize" and the `InitializeParams` datatype as the `params` field.
 
-To authenticate yourself, you must pass in the token in `initializationOptions` as follows:
+If the server is running in TCP mode, to authenticate yourself, you must pass in the token in `initializationOptions` as follows:
```
eed3si9n, line 148: I have the example written for domain socket in https://developer.lightbend.com/blog/2017-11-30-sbt-1-1-0-RC1-sbt-server/

eed3si9n (changes requested):

src/reference/02-DetailTopics/01-Using-sbt/07-sbt-server.md
martijnhoekstra: I split the examples in three cases, leading with domain sockets, then named pipes, then TCP.

src/reference/02-DetailTopics/01-Using-sbt/07-sbt-server.md
martijnhoekstra: copied the netcat session from the blog as an example.

eed3si9n: beautiful.

eed3si9n (approved):

PR #656, opened 2018-04-15
Explain how to prevent the project from being published

raquo:
When my project was using sbt 0.13.15 I had this in my `root` project settings to prevent it from being published: `publish := {}, publishLocal := {}`.

After upgrading my project to 1.1.4 I needed to replace these with `skip in publish`. However, this is not documented anywhere AFAIK, so I had to find it out the hard way.

My knowledge of these settings is limited to what I could infer from https://github.com/sbt/sbt/issues/3136, which is why the text I'm submitting here is not very detailed, but I think it's still better than nothing.


lightbend-cla-validator:
Hi @raquo,

raquo:
Signed the CLA!

eed3si9n:
Nice! Thanks for this.

eed3si9n (approved):

PR #663, opened 2018-05-01
Adds cookie banner to site

rasummer:
@eed3si9n 

Prevents GA from loading if user opts out. Adds standard legal footer, preference center and cookie listing.

If you'd like to test locally, then you need to swap out the One Trust include, template files line 19:
`<script src="https://optanon.blob.core.windows.net/consent/d759c5db-6821-46e0-a988-6bc699efb74e-test.js" type="text/javascript" charset="UTF-8"></script>`

screenshot-localhost-8000-2018 04 30-17-45-16
screenshot-localhost-8000-2018 04 30-17-48-23
screenshot-localhost-8000-2018 04 30-17-48-56

### DIFF
--- a/src/nanoc/content/cookie.md
+++ b/src/nanoc/content/cookie.md
@@ -1,0 +1,11 @@
+---
+title: Cookie
+---
+
+<h1>Your Privacy</h1>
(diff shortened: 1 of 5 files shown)


eed3si9n (approved):
Thanks @rasummer!

PR #670, opened 2018-05-09
sbt 1.1.5

eed3si9n:


src/nanoc/nanoc.yaml
```diff
@@ -1,6 +1,6 @@
-sbtVersion: 1.1.4
+sbtVersion: 1.1.5
 # Windows build version tends to be different from the rest
-windowsBuild: 1.1.4
+windowsBuild: 1.1.5
 
 sbtVersionForScalaDoc: 1.1.2
```
SethTisue: is this an oversight?

eed3si9n: yes. we need a better oversight of these process :)

SethTisue: `git grep 1\\.\\.` ftw

PR #683, opened 2018-05-30
Note where PublishLocal writes

martijnhoekstra:


dwijnand:
Oh.. my change pushed after yours. Feel free to drop it: your second change already resolves my concern. 😄 

Nice work!

dwijnand (approved):

PR #684, opened 2018-06-02
fix some url

xuwei-k:


dwijnand:
Excellent, thanks Yoshida-san.

PR #693, opened 2018-06-20
sourceManaged -> managedSources

kubukoz:


kubukoz:
Forget it, there are actually both :D

PR #704, opened 2018-07-20
Updated plugin best practice

eed3si9n:


src/reference/02-DetailTopics/05-Plugins-and-Best-Practices/03-Plugins-Best-Practices.md
```diff
@@ -270,6 +294,68 @@ lazy val app = (project in file("app"))
...
+but is used as `(obfuscate / obfuscateOption)` in the `projectSettings`.
+This lets the user either set `obfuscate / obfuscateOption` at a particular subproject level,
+or scoped to `ThisBuild` affecting all subprojects:
+
+```scala
+ThisBuild / obfuscate / obfuscateOption := ObfuscateOption().withX(true)
+```
+
+This requires in-depth knowledge to know that RHS are all contained in
+global scope (pure values and other global settings), but it's a useful technique.
```
dwijnand, line 357: What do you mean here? It doesn't make sense to me.

dwijnand:
:+1: very nice changes

src/reference/02-DetailTopics/05-Plugins-and-Best-Practices/03-Plugins-Best-Practices.md
eed3si9n: I am saying that since there's no dynamic dispatch, you the plugin author need to know/guess that all the stuff in RHS are actually safe to call in global scope. For example `compile` is probably not good but what about `libraryDependencies` etc.

dwijnand: Ah, I see.

PR #734, opened 2018-10-18
Tweak cached resolution example to be more concise

rtyley:
Using the `~=` operator is more concise here, right?

### DIFF
--- a/src/reference/02-DetailTopics/03-Dependency-Management/08-Cached-Resolution.md
+++ b/src/reference/02-DetailTopics/03-Dependency-Management/08-Cached-Resolution.md
@@ -15,7 +15,7 @@ Cached resolution is an **experimental** feature of sbt added since 0.13.7 to ad
 To set up cached resolution include the following setting in your project's build:
 
 ```scala
-updateOptions := updateOptions.value.withCachedResolution(true)
+updateOptions ~= { _.withCachedResolution(true) }
 ```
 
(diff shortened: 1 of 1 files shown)


eed3si9n:
In documentation sbt team has been been migrating to using `:=` and `+=` almost exclusively, and as far as I can tell `~=` is not even documented unless you look at sbt 0.12 documentation any more - https://www.scala-sbt.org/0.12.4/docs/Getting-Started/More-About-Settings.html

It's a neat shorthand for advanced users, and probably I should recover its documentation in some detailed topic section, but for cached resolution I think it's better to keep it verbose but easier to understand `:=` + `.value`.

rtyley:
Ah fair enough - so at the moment the status of `~=` is undocumented but not deprecated, is that right?

eed3si9n:
I think so. I just checked using grep:

```
$ rg ~=
src/dormant/Settings.md
115:`<++=`, and `~=` are convenience methods that can be defined in terms of
184:### ~=
186:`~=` is used to transform the current value of a setting. For example,
191:scalacOptions in Compile ~= { (options: Seq[String]) =>
200:libraryDependencies ~= { (deps: Seq[ModuleID]) =>
320:method used. In the case of `:=`, `+=`, `++=`, and `~=`, the type of the

src/dormant/Needs-New-Home.md
149:projectID ~= { id =>
```

PR #744, opened 2018-11-25
Fix up Windows-related first-run examples.

bab9e9:
Touch command does not exist in windows cmd. Prepare the user for the hundreds of lines of messages generated on first run.


bab9e9:
If Windows is supported, do it. Don't just expect users to translate from Linux commands.

PR #749, opened 2018-12-21
Add sbt-stryker4s

hugo-vrijswijk:
I've added Stryker4s, a mutation testing tool for Scala. We have just released our first version of the sbt plugin


eed3si9n (approved):
Thanks!

PR #753, opened 2018-12-30
Add a link to sbt-lit

earldouglas:
This could fall under either documentation plugins, since documentation and literate programming are both focused on communicating with other humans, or code generator plugins, which is the sbt mechanism used to extract code from literate files.


eed3si9n (approved):
Thanks @earldouglas!

PR #759, opened 2019-01-11
Add sbt-delombok

Atry:


eed3si9n (approved):
Thanks @Atry

eed3si9n:
The PR validation broken since Yahoo API stopped working without authentication.

PR #767, opened 2019-02-21
update REPL output to match its input

LeeiFrankJaw:
The section _sbt by example_ was updated but the REPL output did not
match its input and this was confusing.


eed3si9n (approved):
Thanks!

PR #774, opened 2019-03-26
Add a reference to the sbt-autoplugin.g8 template

coreyoconnor:
The sbt-autoplugin.g8 template is a nice template for a new sbt plugin that targets publishing to the community. I didn't see an existing reference and I think this template is worth highlighting early in the instructions.


eed3si9n (approved):
Thanks @coreyoconnor!

PR #786, opened 2019-06-10
Classloader layering

eatkins:


src/reference/02-DetailTopics/02-Configuration/15-In-Process-Classloaders.md
```diff
@@ -0,0 +1,92 @@
...
+intermediate layers.
+
+Another benefit of the layered classloaders is that the jvm may choose to jit
+some of the cached classes. As a result, the total runtime of the task may
+decrease as the interactive session proceeds.
+
+
+### ClassLoader layering strategy
+As of 1.3.0, sbt allows users to select one of three classloader layering
+strategies[*]. The strategy may be set using the `classLoaderLayeringStrategy`
```
eed3si9n: I'm guessing that `[*]` won't work with Pamflet. Could you inline it with parens instead?

src/reference/02-DetailTopics/02-Configuration/15-In-Process-Classloaders.md
```diff
@@ -0,0 +1,92 @@
...
+out: In-Process-Classloaders.html
+---
+
+  [Running Project Code]: Running-Project-Code.html
+  [Forking]: Forking.html
+
+
+### In process class loading
+
+When evaluating the `run` and `test` tasks without [forking][Forking], sbt
```
eed3si9n: Could we start with 

> `run` and `test` tasks by default emulate `java` invocation using function calls with a ClassLoader trick.

then we can continue to

> Unless [forking][Forking] is set, sbt creates a custom ClassLoader for each `run` evaluation.

src/reference/02-DetailTopics/02-Configuration/15-In-Process-Classloaders.md
```diff
@@ -0,0 +1,92 @@
...
+creates a custom classloader for each task evaluation run. For many projects, a
+large amount of time can be spent loading the classes for the task. This is
+especially true for projects with many dependencies because the jvm must find
```
eed3si9n: for consistency, capitalize three letter acronyms like jvm, jar, jit, and ssd please.

src/reference/02-DetailTopics/02-Configuration/15-In-Process-Classloaders.md
```diff
@@ -0,0 +1,92 @@
...
+all of the required classes in a lengthy list of jar files. Even with ssds, the
+io required to locate all of the classes and read them into memory can delay the
+startup of the task by seconds. To reduce this overhead, sbt creates layered
```
eed3si9n: The use of "sbt" as subject muddies the situation a bit since this is no longer available by default.
Maybe

> To reduce this overhead, the build can be configured to use layered ClassLoader where certain layers may be cached between task evaluation runs.

src/reference/02-DetailTopics/02-Configuration/15-In-Process-Classloaders.md
```diff
@@ -0,0 +1,92 @@
...
+
+```
+Test / classLoaderLayeringStrategy := ClassLoaderLayeringStrategy.AllLibraryJars
+```
+
+Descriptions of the strategies follow.
+
+#### ScalaLibrary
+
+Prior to sbt 1.3.0, this was the only ClassLoader layering strategy. The top
```
eed3si9n: The first sentence should summarize the thing, like

> `ClassLoaderLayeringStrategy.ScalaLibrary` creates a two-layer ClassLoader where Scala standard library is kept warm, similar to sbt 1.2.x.

src/reference/02-DetailTopics/02-Configuration/15-In-Process-Classloaders.md
```diff
@@ -0,0 +1,92 @@
...
+layer of this ClassLoader is able to load the scala standard library classes,
```
eed3si9n: > The top layer of this ClassLoader

In terms of what's top/bottom, I think it's more intuitive for most users if the upper layers extend lower layers as typically stacked diagrams are written. This is rather unfortunate since "top ClassLoader" means something, but since we are discussing layering strategy, I propose that we flip up/down here.

> The bottom layer loads the Scala standard library classes, the sbt test interface classes, and the Java library.

src/reference/02-DetailTopics/02-Configuration/15-In-Process-Classloaders.md
```diff
@@ -0,0 +1,92 @@
...
+Note that because the scala library is cached between runs, any shared global
+state created by the scala library is reused. In particular, the
+[global execution context](https://www.scala-lang.org/api/current/scala/concurrent/ExecutionContext\$.html)
+persists between runs. Avoid using the global execution context for tasks that
+run in process.
+
+The `ScalaLibrary` strategy is the default for all sbt 1.3.0 builds.
+
+#### AllLibraryJars
+
```
eed3si9n: From the release note:

> `ClassLoaderLayeringStrategy.AllLibraryJars` creates a three-layer ClassLoader where library dependencies, in addition to Scala standard libraries are kept warm.

src/reference/02-DetailTopics/02-Configuration/15-In-Process-Classloaders.md
```diff
@@ -0,0 +1,92 @@
...
+The requirements for java serialization and deserialization to work are the same
+as that described in [Running Project Code][Running Project Code]. If, however,
+java serialization/deserialization of classes in the bottom layer is performed
+by classes in the dependency layer, then the dependency layer is invalidated
+between runs. When invalidation occurs, the performance of `AllLibraryJars` will
+be largely indistinguishable from `ScalaLibrary.`
+
+#### Flat
+
```
eed3si9n: > `ClassLoaderLayeringStrategy.Flat` includes all classes and JARs except for the Java runtime. The behavior of tasks using this strategy should be similar to forking without the overhead of starting a new JVM.

src/reference/02-DetailTopics/02-Configuration/15-In-Process-Classloaders.md
```diff
@@ -0,0 +1,92 @@
...
+latency of run and test for projects with a large dependency classpath.
+
+Many projects will benefit in interactive sessions from switching to
+`AllLibraryJars`. The reason that it is not the default is because the caching
+of the dependency layer can cause errors if static fields are shared across
+runs. For example, some libraries use a global cache that is stored in a static
+field. If a project's `run` or `test` task depends on a particular initial
+state, they may fail on successive runs. Note that static global fields in the
+_project_ classpath do not cause this problem because the bottom layer is not
+cached between runs.
```
eed3si9n: I'd cut out the 

> For example, some libraries use a global cache that is stored in a static
field. If a project's `run` or `test` task depends on a particular initial
state, they may fail on successive runs. Note that static global fields in the
_project_ classpath do not cause this problem because the bottom layer is not
cached between runs.

since "because the caching of the dependency layer can cause errors if static fields are shared" is a good-enough summary.

Also project vs library could be difficult to distinguish if projects are split into multiple repos.

src/reference/02-DetailTopics/02-Configuration/15-In-Process-Classloaders.md
```diff
@@ -0,0 +1,92 @@
...
+
+The requirements for java serialization and deserialization to work are the same
```
eed3si9n: I'd reword this section to

> **Note**: Java serialization does not work well with ClassLoader layering. If you know Java serialization is used, stick with `ClassLoaderLayeringStrategy.ScalaLibrary` or `ClassLoaderLayeringStrategy.Flat`.

src/reference/02-DetailTopics/02-Configuration/15-In-Process-Classloaders.md
```diff
@@ -0,0 +1,92 @@
...
+
+#### Flat
+
+This has only a single layer. It is essentially a URLClassLoader whose jars are
+the result of `fullClasspath.value.files`. Using this strategy should have
+similar characteristics as forking. The startup latency should be a bit lower
+than forking because it need not create a new jvm, but it has the same
+limitations with respect to shutdown hooks and threading as described in
+[Running Project Code][Running Project Code]. In most cases, `ScalaLibrary` or
+forking should be preferred.
```
eed3si9n: > In most cases, `ScalaLibrary` or forking should be preferred.

Since we've already promoted the goodness of other layering strategies, I don't think this sentence is necessary.
Instead, we should let the users know when `ClassLoaderLayeringStrategy.Flat` might come in handy.

For instance we can say:

> If in previous versions of sbt when forking was needed (due to the use of global ExecutionContext etc), consider trying `ClassLoaderLayeringStrategy.Flat`.

src/reference/02-DetailTopics/02-Configuration/15-In-Process-Classloaders.md
```diff
@@ -0,0 +1,92 @@
...
+as that described in [Running Project Code][Running Project Code]. If, however,
+java serialization/deserialization of classes in the bottom layer is performed
+by classes in the dependency layer, then the dependency layer is invalidated
+between runs. When invalidation occurs, the performance of `AllLibraryJars` will
+be largely indistinguishable from `ScalaLibrary.`
+
+#### Flat
+
+This has only a single layer. It is essentially a URLClassLoader whose jars are
+the result of `fullClasspath.value.files`. Using this strategy should have
```
eed3si9n: > It is essentially a URLClassLoader whose jars are the result of `fullClasspath.value.files`.

I'd cut this one out.

eed3si9n (changes requested):
some suggestions.

lightbend-cla-validator:
At least one pull request committer is not linked to a user. See https://help.github.com/en/articles/why-are-my-commits-linked-to-the-wrong-user#commits-are-not-linked-to-any-user

eatkins:
@eed3si9n I ended up largely rewriting this with many of your suggestions in mind.

eed3si9n (approved):

eed3si9n:
LGTM

PR #787, opened 2019-06-10
Document pandoc version and update pandoc

### LINKED ISSUE #564: Specific version of pandoc.
Since `pandoc 2.0.0`, it renamed `--latex-engine` to `--pdf-engine`. There is [the release note of pandoc 2.0.0](https://pandoc.org/releases.html#behavior-changes). I guess it would be better to indicate which version we use. Or maybe we could update argument of `pandoc`, in the meantime, we could document the version we use.

jiminhsieh:
Fix #564 and upgrade pandoc to 2.3.1.

The latest of pandoc is 2.7.2, but [there are behavior changes at 2.4](https://pandoc.org/releases.html#pandoc-2.4-3-november-2018). To prevent the behavior changes, I used the version before 2.4 which is 2.3.1.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ scala-sbt.org is powered by:
 
 * [Pamflet](http://www.foundweekends.org/pamflet/), a Scala-based documentation engine written by [@n8han][] (and some contributions from [@eed3si9n][]) generates the sbt 0.13/1.x documentation.
 
-* [Pandoc](https://pandoc.org/), to generate pdf files.
+* [Pandoc](https://pandoc.org/) [2.3.1](https://pandoc.org/releases.html#pandoc-2.3.1-28-september-2018), to generate pdf files.
 
 [@n8han]: https://github.com/n8han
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -9,7 +9,7 @@ scala-sbt.org is powered by:
 
 * [Pamflet](http://www.foundweekends.org/pamflet/), a Scala-based documentation engine written by [@n8han][] (and some contributions from [@eed3si9n][]) generates the sbt 0.13/1.x documentation.
 
-* [Pandoc](https://pandoc.org/), to generate pdf files.
+* [Pandoc](https://pandoc.org/) [2.3.1](https://pandoc.org/releases.html#pandoc-2.3.1-28-september-2018), to generate pdf files.
```
jiminhsieh, line 12: Maybe there is a better place to indicate which version we used? 🤔

eed3si9n (approved):
Thanks!

PR #795, opened 2019-07-19
Watch fixes

eatkins:


eed3si9n (approved):
The changes looks fine.
It might be good to also mention the newly added keyboard shortcuts?

eatkins:
I thought about that. The reason I left it out is that the shortcuts are self-documenting with the '?' option. I'm not opposed to adding documentation here as well, but it felt redundant.

eed3si9n:
It probably is redundant, but there's a benefit to having the info on the html so it gets indexed by Google, etc.

PR #798, opened 2019-07-31
Recommend downloading sbt's key securely

cb372:
Update the instructions for Debian installation to download the public key over TLS


eed3si9n (approved):
Thanks @cb372!

PR #800, opened 2019-07-31
fix: use syntax highlighting for code snippet in Datatype

hugo-vrijswijk:
Small fix to make the code snippet more readable.


eed3si9n (approved):
Thanks @hugo-vrijswijk!

PR #809, opened 2019-08-28
Document ScopeFilter / .all's val requirement

dwijnand:


eed3si9n (approved):
Haven't confirmed it, but I'm merging this.

PR #823, opened 2019-09-21
Add sbt-no-publish

ChristopherDavenport:
Simple Plugin for disabling project publication. This keeps getting duplicated everywhere so making a plugin to turn it off made sense to me.

Just enable explicitly to turn off publishing. `.enablePlugin(NoPublishPlugin)


eed3si9n:
Doesn't `publish / skip := true` work?

ChristopherDavenport:
That may be the case, and this was then unnecessary(but I learned something valuable). All the builds I encounter generally set.

```
skip in publish := true,
publish := (()),
publishLocal := (()),
publishArtifact := false,
publishTo := None
```

If these are unnecessary. That's great! But there's a lack of information getting around the ecosystem effectively.

eed3si9n:
Yea. Looks like it was one the features that shipped with sbt 1.0 but wasn't communicated well.

https://github.com/sbt/sbt/pull/3380

ChristopherDavenport:
Great! Anywhere I was going to use this, I'll use that instead.

ReoMaterSom:
#836

eed3si9n:
I added documentation of skipping as part of https://github.com/sbt/website/pull/832

PR #847, opened 2019-10-29
Fix shell code for download on linux

felher:
The code to paste into the shell on the official download page for linux has some problems (at least on my machine)

- It uses unicode quotation marks instead of regular ones
- It doesn't preserve newlines

This leads to no (or bad) results when pasting it into the shell. I think the problems are added automatically by the markdown processor.

This commit substitutes the \<pre\> tag, which was initially used to display the shell lines, to code \`\`\` delimiters, which seems to fix the issue.

### DIFF
--- a/src/landing/download.md
+++ b/src/landing/download.md
@@ -41,21 +41,21 @@ Windows
 
   <div class="distro_debian">
   	<h2>Linux (deb)</h2>
-<pre>
+```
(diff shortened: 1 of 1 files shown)


lightbend-cla-validator:
At least one pull request committer is not linked to a user. See https://help.github.com/en/articles/why-are-my-commits-linked-to-the-wrong-user#commits-are-not-linked-to-any-user

eed3si9n (approved):
Thanks

PR #859, opened 2019-12-05
clarify color for `sbt.ci`

francisdb:


francisdb:
After some more testing it looks like `-Dsbt.ci=true` does not touch any color settings?
see https://github.com/sbt/sbt/issues/5269

PR #863, opened 2019-12-25
Update 11-Custom-Settings.md

fancywriter:
fixing small typo in documentation

### DIFF
--- a/src/reference/00-Getting-Started/11-Custom-Settings.md
+++ b/src/reference/00-Getting-Started/11-Custom-Settings.md
@@ -230,7 +230,7 @@ lazy val library = (project in file("library"))
       println("s: " + s)
       s
     },
-    sampleStringTask := {
+    stopServer := {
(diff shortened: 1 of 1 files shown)


src/reference/00-Getting-Started/11-Custom-Settings.md
```diff
@@ -230,7 +230,7 @@ lazy val library = (project in file("library"))
       println("s: " + s)
       s
     },
-    sampleStringTask := {
+    stopServer := {
```
eed3si9n, line 233: This is not a typo. See in the following `sampleStringTask.value` is assigned to `val old`: If this is called `stopServer` calling it `old` won't make sense. In `build.sbt` you can rewire keys while referring to the old behavior, similar to class inheritance and `super`.

fancywriter: 😮  oh, will reread it again. thank you.

PR #864, opened 2020-01-04
Travis CI: update the sample

dwijnand:
Sync the sample with the new instructions.


eed3si9n (changes requested):
`$` needs to be escaped.

eed3si9n (approved):

PR #882, opened 2020-03-18
Fix StateTransform description/example [sbt/sbt#5477]

dcsobral:


dcsobral:
Fixes sbt/sbt#5477.

dwijnand:
I understand from https://github.com/sbt/sbt/issues/5477#issuecomment-600945279 that this was the 1.3.x API, so by now this doc change would make it worse.

PR #908, opened 2020-07-17
Make tables striped so they will be easier to read

marcospereira:
In the current form, it is sometimes hard to tell which text belongs to each row. Making the tables striped makes it clear.

## Before

Screen Shot 2020-07-17 at 1 58 31 PM

## After

Screen Shot 2020-07-17 at 1 58 13 PM

---

Also, since I was here, fix some (low-hanging) typos.

### DIFF
--- a/src/reference/01-General-Info/07-Contributing-to-sbt.md
+++ b/src/reference/01-General-Info/07-Contributing-to-sbt.md
@@ -1,5 +1,5 @@
 ---
-out: Contibuting-to-sbt.html
+out: Contributing-to-sbt.html
 ---
 
   [sbt-dev]: https://groups.google.com/d/forum/sbt-dev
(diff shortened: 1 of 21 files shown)


src/reference/01-General-Info/07-Contributing-to-sbt.md
```diff
@@ -1,5 +1,5 @@
 ---
-out: Contibuting-to-sbt.html
```
marcospereira, line 2: Not sure about this one, since it will eventually break some links in the wild wild web.

eed3si9n: I think it's ok.

eed3si9n (approved):
Thanks @marcospereira!

PR #913, opened 2020-08-21
Add sbt-rewarn to the list of community plugins

rtimush:
I'm not sure if it should go to the "Utility and system plugins" or to the "Static code analysis plugins" sections.


eed3si9n:
sbt-errors-summary is in Utility and system plugins so probably the same as that?

eed3si9n (approved):

PR #919, opened 2020-09-19
Add information how to get a Logger for running a process

ScalaWilliam:
It was missing - as a long time user of SBT I first tried to do `val log = streams.value` which did not work.
Took a few Google searches to find the right one.


eed3si9n (approved):
Thanks!

PR #948, opened 2020-11-07
Add instructions for installing sbtn

eatkins:


src/reference/00-Getting-Started/01-Setup/d.md
```diff
@@ -0,0 +1,126 @@
...
+
+Binary releases of sbtn are available at
+[sbtn-dist](https://github.com/sbt/sbtn-dist/tags). The latest platform specific
+releases can be found at:
+
+* Linux <https://github.com/sbt/sbtn-dist/releases/download/v$app_version$/sbtn-x86_64-pc-linux-$app_version$.tar.gz>
+
+* MacOS <https://github.com/sbt/sbtn-dist/releases/download/v$app_version$/sbtn-x86_64-apple-darwin-$app_version$.tar.gz>
+
+* Windows <https://github.com/sbt/sbtn-dist/releases/download/v$app_version$/sbtn-x86_64-pc-win32-$app_version$.zip>
```
eed3si9n, line 62: sbtn version doesn't always correspond with sbt version, since I was hoping that we wouldn't have to update it once it's stable. For example sbt 1.4.1 uses sbtn 1.4.0, I think.

eed3si9n:
I think we should also add the information about `sbt --client` here. I can add that information later.

src/reference/00-Getting-Started/01-Setup/d.md
eatkins: I wondered about that. The reason I put the app_version in the url is because I wanted the download urls to be available and reflect the latest code. I would suggest that until sbtn is truly stable that we keep releasing sbtn versions along with sbt versions even if nothing has changed. It makes it easier to think about versions if they are kept in lockstep both for documentation and troubleshooting. I can think of a few alternatives though:
1. Provide urls that always redirect to the latest version of sbtn
2. Add an` sbtn_version` to the project templates and make sure that we update it along with app_version whenever making a new release (independent of sbtn installation instructions, it seems like bumping app_version should be added to the release process).

I would prefer Option 1 to what I did in this PR but I don't know how to implement it.

eatkins:
One thing about `--client`: it's a little confusing that `-client` is different from `--client`. Moreover, it's unfortunate that `--client` is generally a better experience than `-client`, but `-client` takes one less character to type. I'm wondering if we can change `-client` maybe to be the same as `--client` and make a new argument that's like `--java-client` or something for the old behavior currently implemented by `-client`.

eed3si9n:
> One thing about `--client`: it's a little confusing that `-client` is different from `--client`.

I'd consider this a bug. Given `sbtn` the old thin client should be used only as a fallback client on BSD or Debian (because apt get doesn't include sbtn any more to save bandwidth).

leifwickland:
These directions worked seamlessly for me on Mac OS.
Thanks so much for making this dead easy.

PR #970, opened 2020-12-15
Fix branch syntax for sbt new

jrudolph:
It seems `--branch=<branch>` does not work but without the `=` it does.

Otherwise, you'll get this error:

```
Ignoring unrecognized parameter: branch
```


eed3si9n (approved):
Thanks!

jrudolph:
I tried to find where that restriction not to use `=` might come from. It seems it's from giter8 itself which uses `scopt` to parse arguments which just might not support `=` at all in that form?

PR #971, opened 2020-12-19
Update plugin documentation

eed3si9n:
Fixes https://github.com/sbt/sbt/issues/6131

PR #972, opened 2020-12-20
Some follow-up explanation about plugins

eed3si9n:
Ref https://github.com/sbt/website/pull/971

/cc @logc


logc:
Thanks for addressing my comments on #971 , @eed3si9n , and sorry for my late replies! I am on holidays and away from my usual development environment.

PR #973, opened 2020-12-29
Update xsbt-web-plugin description

earldouglas:
This updates the description and stars count for xsbt-web-plugin to
match the upstream project.


eed3si9n (approved):
Thanks!

PR #977, opened 2021-01-06
clarify what DSL stands for

vivian-duong:
What DSL stands for is not obvious and is only expanded much later in this Getting Started Guide.


eed3si9n (approved):
Thanks!

PR #985, opened 2021-02-08
How to define a custom dependency configuration

eed3si9n:


src/reference/04-Howto/30-Howto-Custom-Dependency-Configuration.md
```diff
@@ -0,0 +1,88 @@
...
+
+Some configurations you'll see in sbt:
+
+- `Compile` which defines the main build (`src/main/scala`).
+- `Test` which defines how to build tests (`src/test/scala`).
+- `Runtime` which defines the classpath for the `run` task.
+
+### Cautions on custom dependency configurations
+
+A custom configuration shoul be considered only when you are
```
bjaglin:
```suggestion
A custom configuration should be considered only when you are
```

bjaglin (commented):
Thank you @eed3si9n ! Have you considered referencing https://www.scala-sbt.org/1.x/docs/Advanced-Configurations-Example.html?

I know you rightly mention that custom configurations bring complexity, but going back to https://github.com/sbt/sbt/pull/6296#discussion_r571648724, as a reader I am still unclear about whether `IntegrationTest` is a custom configuration or not, and how it affects usage of sbt keys packaged as plugins.

Particularly, I need to know that semanticdb functionnality is packaged as a plugin and thus
```scala
.configs(IntegrationTest)
.settings(Defaults.itSettings),
```
is not enough to get [proper](https://github.com/sbt/sbt/pull/6296#discussion_r571502345) support of semanticdb.

From a user perspective and ignoring backward compatibility concerns, I would say that either
- the fact that semanticdb keys are exposed by an internal `allRequirements` plugin is an implementation detail of sbt, and thus `Defaults.itSettings` should source `sbt.plugins.SemanticdbPlugin.configurationSettings`
- the `SemanticdbPlugin` needs to be explicitly activated (`noTrigger`), signaling the fact that `sbt.plugins.SemanticdbPlugin.configurationSettings` must be sourced manually for `IntegrationTest` (like any other third-party plugin).

WDYT?

eed3si9n:
@bjaglin 

> Have you considered referencing https://www.scala-sbt.org/1.x/docs/Advanced-Configurations-Example.html?

I've looked at it a few times in the context of writing this, but I can't in good conscience expose more people to that as example, because I don't think people should write build files like that, especially about the bits on overriding `Compile` sounds risky.

> I am still unclear about whether `IntegrationTest` is a custom configuration or not, and how it affects usage of sbt keys packaged as plugins.

Since it's widely used, I think we should consider `IntegrationTest` to be a semi-official configuration and it should be as user-friendly as we can make it.

> the fact that semanticdb keys are exposed by an internal `allRequirements` plugin is an implementation detail of sbt, and thus `Defaults.itSettings` should source `sbt.plugins.SemanticdbPlugin.configurationSettings`

Or maybe `configSettings`?

https://github.com/sbt/sbt/blob/8d8a37675798d9e714755bbe0bce91bc141f1a85/main/src/main/scala/sbt/Defaults.scala#L2492-L2495

bjaglin:
> > the fact that semanticdb keys are exposed by an internal `allRequirements` plugin is an implementation detail of sbt, and thus `Defaults.itSettings` should source `sbt.plugins.SemanticdbPlugin.configurationSettings`
> 
> Or maybe `configSettings`?
> 
> https://github.com/sbt/sbt/blob/8d8a37675798d9e714755bbe0bce91bc141f1a85/main/src/main/scala/sbt/Defaults.scala#L2492-L2495

Indeed. I'll push an extra commit on https://github.com/sbt/sbt/pull/6296 shortly.

PR #998, opened 2021-03-09
Add sbt-scala2plantuml plugin

steinybot:
I wasn't too sure on the line length rule being used here.


eed3si9n (approved):

eed3si9n:
> I wasn't too sure on the line length rule being used here.

Whatever works I guess.

PR #1002, opened 2021-03-12
Include more detail about the assumptions of the shell

ScalaWilliam:
Especially for Windows users.

### DIFF
--- a/src/reference/00-Getting-Started/02-sbt-by-example.md
+++ b/src/reference/00-Getting-Started/02-sbt-by-example.md
@@ -10,7 +10,7 @@ out: sbt-by-example.html
 sbt by example
 --------------
 
-This page assumes you've [installed sbt 1][Setup].
+This page assumes you've [installed sbt 1][Setup] and you are running a Bash/UNIX shell on macOS, Linux or WSL ([Windows Subsystem for Linux](https://docs.microsoft.com/en-us/windows/wsl/install-win10)).
 
 Let's start with examples rather than explaining how sbt works or why.
 


src/reference/00-Getting-Started/02-sbt-by-example.md
```diff
@@ -10,7 +10,7 @@ out: sbt-by-example.html
 sbt by example
 --------------
 
-This page assumes you've [installed sbt 1][Setup].
+This page assumes you've [installed sbt 1][Setup] and you are running a Bash/UNIX shell on macOS, Linux or WSL ([Windows Subsystem for Linux](https://docs.microsoft.com/en-us/windows/wsl/install-win10)).
```
eed3si9n, line 13: Why does the system shell matter so much here? Are we trying to preclude Windows console?

ScalaWilliam: These instructions would not work on the Windows default console, and some new users might be just confused what this all means and skip the tutorial.

ScalaWilliam: And many people still don't know about WSL being available on Windows at all.

ScalaWilliam: for example:
image

eed3si9n: If the user is already familiar with Windows console, I don't want them to switch to using WSL.
I'd much rather document touch equivalent on console together, which is

```
type nul >> build.sbt
```

?

PR #1005, opened 2021-03-21
Inform the user how to find & verify the Coursier cache location

ScalaWilliam:
Related to https://github.com/sbt/sbt/issues/5120

The `csr` key was not so obvious so should be mentioned in the docs.


eed3si9n (approved):
Thanks @ScalaWilliam

PR #1008, opened 2021-03-22
Add documentation for Scala 3 sandwiches

### LINKED ISSUE #1007: Documentation for the Scala 2.13-3.x sandwich
Currently the behavior of the Scala 2<->3 dependencies and then new methods such as `CrossVersion.for3Use2_13` are only documented in the release notes: https://eed3si9n.com/sbt-1.5.0-beta, ideally they should be part of https://www.scala-sbt.org/1.x/docs/Cross-Build.html so we have one complete resource we can point people to /cc @adpi2

adpi2:
Fixes #1007 
@smarter

### DIFF
--- a/src/reference/02-DetailTopics/01-Using-sbt/03-Cross-Build.md
+++ b/src/reference/02-DetailTopics/01-Using-sbt/03-Cross-Build.md
@@ -285,6 +285,29 @@ depending on the value of `crossPaths`.
 Because (unlike Scala library) Scala compiler is not forward compatible among
 the patch releases, compiler plugins should use `CrossVersion.full`.
 
+#### Scala 3 specific cross-versions 
+
+In a Scala 3 project you can use Scala 2.13 libraries:
+
+```scala
+("a" % "b" % "1.0") cross CrossVersion.for3Use2_13
+```
+
+This is equivalent to using `%%` except it resolves the `_2.13` variant 
+of the library  when `scalaVersion` is 3.x.y.
+
+Conversely we have `CrossVersion.for2_13Use3` to use the `_3` variant of the
+library when `scalaVersion` is 2.13.x:
+
+```scala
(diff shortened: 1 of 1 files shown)


src/reference/02-DetailTopics/01-Using-sbt/03-Cross-Build.md
```diff
@@ -341,6 +341,23 @@ than the default.
   }
 }
 ```
+### Scala 3 specific cross-versions
```
smarter: Thanks! Since Scala 3 cross-building is likely to be used more than the other CrossVersion stuff, maybe this should be near the top of page, or maybe somewhere at the top of the page there can be a link to this section?

eed3si9n: In the release note I added a chunky warning with `CrossVersion.for3Use2_13`. I think we should bring that in as well since it's not risk-free to do this.

smarter: Agreed, though I think that warning as-is looks more scary than it is: it sounds like application developers need to carefully inspect their transitive dependencies to avoid mixing _2.13 and _3, but actually sbt detects these conflicts itself and emits an error, so it shouldn't be possible to accidentally end up with both versions on the classpath.

eed3si9n: Not sure what % is app developers vs what % is library authors, but generally it's safe for application developers (or in-house library authors), but we should encourage library authors to cross build against 3.0 universe when possible and avoid tangling 2.13 and/or 3.x back and forth (as djspiewak advocates). Generally there are more app developers than library authors, but if you add "someone who is reading sbt cross building docs" as a prior, maybe the percentage is 50:50?

eed3si9n (approved):

PR #1014, opened 2021-04-11
Scalatest version increased

xhudik:
some example couldn't be compiled - I fixed it. Then I have realized that it was because I used a newer Scalatest version.  With the original Scalatest version - all is working. However, in the meantime I increased Scalatest version. 
I'm not sure whether this is useful. Feel free to cancel


eed3si9n (approved):
Thanks!

PR #1036, opened 2021-09-04
Using Sonatype: Replace deprecated sks-keyserver example with other keyserver

todokr:
[sks-keyserver](https://sks-keyservers.net/) is deprecated and no longer maintained.  

As an alternative, how about replacing the sks one with `keyserver.ubuntu.com` which is mentioned in [The Central Repository doc](https://central.sonatype.org/publish/requirements/gpg/#distributing-your-public-key)?


eed3si9n (approved):
Yes. Thanks for the update.

PR #1038, opened 2021-09-07
Remove the x-overflow: hidden to get the search box overflow working …

er1c:
…again

The search box still shows up if you make the page narrow, but I think that's ok


eed3si9n (approved):
Thanks!

PR #1060, opened 2022-01-22
add explicit `main` instead of deprecated `scala.App`

xuwei-k:
https://docs.scala-lang.org/scala3/book/methods-main-methods.html

> The previous functionality of App, which relied on the “magic” DelayedInit trait, is no longer available. App still exists in limited form for now, but it doesn’t support command line arguments and will be deprecated in the future.


eed3si9n (approved):
LGTM

PR #1073, opened 2022-03-27
use `blackbox.Context`. avoid deprecated Context

xuwei-k:
https://github.com/scala/scala/blob/v2.13.8/src/reflect/scala/reflect/macros/package.scala#L38-L39


eed3si9n (approved):
Thanks!

PR #1103, opened 2022-08-13
Adds link to Inspecting-Settings

sv3ndk:
There's a "see Inspecting-Settings" in plain text in the sentence, whereas I'm assuming the intent is to have a link (as in the next sentence).

Also, for some reason the `[.sbt build definition][Basic-Def]` does not seem to be rendered properly in [the online doc](https://www.scala-sbt.org/1.x/docs/Command-Line-Reference.html) just one line above, although I can't figure out why :thinking:

### DIFF
--- a/src/reference/02-DetailTopics/01-Using-sbt/01-Command-Line-Reference.md
+++ b/src/reference/02-DetailTopics/01-Using-sbt/01-Command-Line-Reference.md
@@ -174,10 +174,10 @@ configuration that can be run using a `Test/` prefix.
     reloaded, or the setting is overridden by another set command or
     removed by the session command. See
     [.sbt build definition][Basic-Def] and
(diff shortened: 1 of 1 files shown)


eed3si9n:
> Also, for some reason the `[.sbt build definition][Basic-Def]` does not seem to be rendered properly in [the online doc](https://www.scala-sbt.org/1.x/docs/Command-Line-Reference.html) just one line above, although I can't figure out why 🤔

Basic-Def missing from here:

https://github.com/sbt/website/blob/a12a1a024b2685a09977db5126f562c1c4ae8607/src/reference/02-DetailTopics/01-Using-sbt/01-Command-Line-Reference.md?plain=1#L5-L15

eed3si9n (approved):
Thanks!

PR #1104, opened 2022-08-13
Mention that  sbt-projectmatrix is optional

sv3ndk:
As a newcomer to cross-building, I had to go back and forth a few times to figure out if the plugin mentioned here was required. 
I suggest to make this mention explicit to make the doc easier to read


eed3si9n (approved):
Thanks @sv3ndk!

PR #1105, opened 2022-08-14
Configure sbt without editing its launch script

sv3ndk:
The current documentation describes how to pass system properties or JVM options directly as `java` command line arguments, which, if I get things correctly, seems to suggest the user should edit the `sbt` script directly. Since those arguments can now  be specified at the `sbt` command line, I suggest to instruct the user to do that instead. 

Also I couldn't find any mention in the documentation of the mechanism related to `SBT_OPTS` and others, so I figured here would be a good location to describe them?


eed3si9n (approved):
Thanks!

PR #1106, opened 2022-08-16
Update 01-Command-Line-Reference.md

sv3ndk:
As a follow-up of the discussion in, https://github.com/sbt/website/pull/1103 , this PR fixes the  broken link visible in the 2nd bullet point of:
https://www.scala-sbt.org/1.x/docs/Command-Line-Reference.html#Commands+for+managing+the+build+definition

Also, the link for explaining how to set up system properties currently points to Sbt-Launcher.html , which is in the sbt Developer's guide with status WIP. 
IMHO the Setup-Notes.html section now contains more directly usable information for the end user so I suggest direct people there instead. 

Going further, maybe the whole "Command Line Options" section should be moved out of this page and into Setup-Notes.html page, since some system properties and other configuration options are also mentioned there, in order to try to put them all in the same place?


eed3si9n:
> Going further, maybe the whole "Command Line Options" section should be moved out of this page and into Setup-Notes.html page, since some system properties and other configuration options are also mentioned there, in order to try to put them all in the same place?

If we're trying to unify all things command line, I think `01-Command-Line-Reference.md` probably should be the page to contain everything in detail.

sv3ndk:
Thanks for the feed-back. 

My understanding is that the expression "Command Line" refers to two different things within this page: in the top part it refers to  commands that can be entered at the sbt prompt, whereas in the second part  it refers to the `sbt` command itself, as passed to the OS shell. The "Command lines options" of the second part are not options of the commands listed in the "Command line reference" top part. 

How about we keep two separate pages then? Something like:

* the whole content of "setup notes" + the "command line options", i.e. all things related to the `sbt` script and how to influence its behavior with argument or env variables. We could call it "sbt startup parameters" or so 
* the content of the top of this page, which maybe we could rename to "sbt prompt command reference" or so  to remove the ambiguity related to "command line".

We could also move those two pages one after the other in the "4. Detailed topic / a. Using sbt" section, instead of having one in "General information" and the other in "Using sbt" as currently. 

What do you think?

eed3si9n:
Historically sbt project only distributed the sbt-launch.jar and main artifacts, and users were told to write their own shell script to their liking. The existing documentation, including Setup notes, reflects that archaic practice. For the last 10 years or so sbt project has been distributing `sbt` shell script, and some of the modern features like `--client` are implemented in part in the script, so we should treat `sbt` to be the premise of installation, and not make any distinction between the script (also called "runner") and the JAR main (mothership).

Especially considering the state of the page prior to https://github.com/sbt/website/pull/1105, it probably makes sense to just remove the page entirely since those `java ...` were advices on how to write the script.

sv3ndk:
Ah ok, got it, thanks for the explanation, I ignored about the history. Maybe indeed it's best to remove the Setup Notes page entirely then. 

Should I do that as part of the current PR? As well as maybe moving the "sbt JVM options and system properties" and "sbt JVM heap, permgen, and stack sizes" sections of "Setup Notes" here as well? I think keeping some sort doc on how to tune the JVM memory is still relevant today (OOMs during my build earlier this week is what triggered me looking for a solution in the doc).

eed3si9n:
> Should I do that as part of the current PR? As well as maybe moving the "sbt JVM options and system properties" and "sbt JVM heap, permgen, and stack sizes" sections of "Setup Notes" here as well?

It's up to you. The current PR seems like a fine change to merge as-is, but if you want to work on it, I'd be happy to keep it open.

> I think keeping some sort doc on how to tune the JVM memory is still relevant today (OOMs during my build earlier this week is what triggered me looking for a solution in the doc).

Yea. I think that can be a subsection in the "command line reference" unless we find a better home.

lightbend-cla-validator:
At least one commit author (svend.vanderveken@gmail.com) is not linked to a user. See https://help.github.com/en/articles/why-are-my-commits-linked-to-the-wrong-user#commits-are-not-linked-to-any-user

lightbend-cla-validator:
At least one commit author (svend.vanderveken@gmail.com, svend.vanderveken@gmail.com) is not linked to a user. See https://help.github.com/en/articles/why-are-my-commits-linked-to-the-wrong-user#commits-are-not-linked-to-any-user

sv3ndk:
Sorry for the longer delay. 

I moved some of the parts of Setup-Notes into Command-Line-Reference and I removed the Setup-Notes page. 

I generated it locally with `sbt makeSite` and all links seems to still work.

src/reference/02-DetailTopics/01-Using-sbt/01-Command-Line-Reference.md
```diff
@@ -12,9 +13,8 @@ out: Command-Line-Reference.html
   [Commands]: Commands.html
   [Running-Project-Code]: Running-Project-Code.html
   [Testing]: Testing.html
-  [Sbt-Launcher]: Sbt-Launcher.html
 
-Command Line Reference
+Sbt prompt command Line Reference
```
eed3si9n: This page I think is sort of written in the style of a manpage for sbt, so I think Command Line Reference sort of makes sense as the title.

eed3si9n (approved):
Thanks!

PR #1108, opened 2022-08-19
Update SDKMAN installation instructions

marc0der:
The current documentation has the following to determine the Java version when installing through SDKMAN:
```
sdk list java | grep -o "\b8\.[0-9]*\.[0-9]*\-tem" | head -1  
```
When running this, it results in the oldest Temurin version available, currently `8.0.345-tem`. Is this really what we want?

If you do a simple `sdk install java` with no qualifier, it will bring down the LTS version of Temurin which is currently set to `17.0.4-tem`.

### DIFF
--- a/src/includes/install.sh
+++ b/src/includes/install.sh
@@ -1,2 +1,2 @@
-\$ sdk install java \$(sdk list java | grep -o "\b8\.[0-9]*\.[0-9]*\-tem" | head -1)
+\$ sdk install java
 \$ sdk install sbt


eed3si9n:
Scala toolchain and library ecosystem haven't always been tested against all JDK versions all the time, so generally it's advisable to be aware of the JDK version you install.

Maybe in the recent years I'd feel more comfortable recommending JDK 11 as default. Play framework for example I think supports 8 and 11. So if we made someone install 17, that user may run into issues later. Same goes for Scala 2.11.

julienrf:
I think it is time to revisit this since more and more libraries do not support Java 8 anymore.

CharlieStras:
> Scala toolchain and library ecosystem haven't always been tested against all JDK versions all the time, so generally it's advisable to be aware of the JDK version you install.
> 
> Maybe in the recent years I'd feel more comfortable recommending JDK 11 as default. Play framework for example I think supports 8 and 11. So if we made someone install 17, that user may run into issues later. Same goes for Scala 2.11.

There is no Temurin version's JDK 8 anymore in sdkman's list. What I get now are

8.0.\*\*\*-amzn
8.0.\*\*\*-amzn
8.0.\*\*\*-librca
8.0.\*\*\*-kona
8.0.\*\*\*-zulu

Which one should the user pick? Or any one is ok?

eed3si9n:
Please consult https://docs.scala-lang.org/overviews/jdk-compatibility/overview.html for the general compatibility situation around the Scala ecosystem.

If you're a user of Scala language, as opposed to library authors, using [Eclipse Temurin](https://adoptium.net/temurin/releases/) 17 or 21, or Zulu if you're on M1 Macs should be ok. Other JVMs might also work, but we do not test them so I can't guarantee that they function ok. Older JDK 8 is widely tested, so all versions of Scala or sbt would work. Using newer JDK means that languages and tooling releases prior to 2017 might not work.

CharlieStras:
> Please consult https://docs.scala-lang.org/overviews/jdk-compatibility/overview.html for the general compatibility situation around the Scala ecosystem.
> 
> If you're a user of Scala language, as opposed to library authors, using [Eclipse Temurin](https://adoptium.net/temurin/releases/) 17 or 21, or Zulu if you're on M1 Macs should be ok. Other JVMs might also work, but we do not test them so I can't guarantee that they function ok. Older JDK 8 is widely tested, so all versions of Scala or sbt would work. Using newer JDK means that languages and tooling releases prior to 2017 might not work.

Thanks for your reply, I'll check it.

For this documentation [Installing-sbt-on-Mac](https://www.scala-sbt.org/1.x/docs/Installing-sbt-on-Mac.html), can we change it to something like to prevent some misleading.

```bash
$ sdk install java $(sdk list java | grep -o "\b8\.[0-9]*\.[0-9]*\-.*" | head -1)
$ sdk install sbt

PR #1131, opened 2023-04-30
Remove TOFU in `download`

yoshinorin:
It may display correctly on Mac, but it will be TOFU on Windows.

tofu


eed3si9n (approved):
Thanks!

PR #1145, opened 2023-06-20
Remove mentions of Lightbend support, switch to Plausible analytics

julienrf:
image

image


julienrf:
I noticed that the API documentation was not built. I guess it is produced separately?

src/landing/_template/error.st
```diff
@@ -12,15 +12,7 @@
...
-
-    <!-- OneTrust Cookies Consent Notice (Production Standard, scala-sbt.org, en-GB) start -->
-    <script src="https://optanon.blob.core.windows.net/consent/d759c5db-6821-46e0-a988-6bc699efb74e.js" type="text/javascript" charset="UTF-8"></script>
-    <script type="text/javascript">
-      function OptanonWrapper() {
-        //callback when OneTrust is loaded and ready
-      }
-    </script>
-    <!-- OneTrust Cookies Consent Notice (Production Standard, scala-sbt.org, en-GB) end -->
+    <script defer data-domain="scala-sbt.org" src="https://plausible.scala-lang.org/js/script.js"></script>
```
eed3si9n, line 15: Could you explain what this does please?

eed3si9n:
> I noticed that the API documentation was not built. I guess it is produced separately?

Yes, it's built using sbt/sbt's special branch and sbt-unidoc because we have multiple repos.

src/landing/_template/error.st
julienrf: It removes the banner “This site uses cookies to offer you a better browsing experience…”, and loads the Plausible analytics scripts hosted on the scala-lang.org servers. Plausible is an analytics platform that does not use cookies. Maybe cookies are used by another component of the website?

eed3si9n (approved):
LGTM pending CI

PR #1153, opened 2023-08-04
refactor: update sbt-by-example

ckipp01:
I love this guide. I always recommend it to new people getting starting with sbt. The last couple times I did however they always came back to me ab out certain things not working anymore, mainly the console examples. I decided to just go through the whole thing myself and update all of the examples using a more modern sbt version and updating the output to match. There is also a couple other choices that I made while updating things like:

- Changing the weather site we use as an example since it no longer is active
- Changing the libraries being used from gigahorse to sttp and Play JSON to uJson. I didn't do this _just because_ but rather to more align with the libraries that are being used by the [scala toolkit](https://docs.scala-lang.org/toolkit/introduction.html). Hopefully people will be a bit more familiar with these as I don't assume a newcomer to Scala will be using gigahorse or really Play JSON _unless_ they are using Play!. If this isn't ok, let me know and I can try to recreate it using the old libraries, but I think these examples are clearer.


src/reference/00-Getting-Started/02-sbt-by-example.md
```diff
@@ -211,35 +226,34 @@ sbt:Hello> ~testQuick
 
 ### Write a test
 
-Leaving the previous command running, create a file named `src/test/scala/HelloSpec.scala`
+Leaving the previous command running, create a file named `src/test/scala/example/HelloSpec.scala`
```
ckipp01: Tiny change, but I noticed it while going through this. Since we had the user create `src/main/scala/example/` I thought it appropriate to also do the same for the test, which is more than likely going to be followed on their teams anyways.

ckipp01:
Running the `ref/example-weather` I get:

```
[info] + ref/example-weather [PENDING]
[error]  Pending test passed. Mark as passing to remove this failure.
[error] stack trace is suppressed; run last scripted for the full output
[error] (scripted) Failed tests:
[error]         ref/example-weather
```
I don't see any errors though when it runs so I don't fully get what is failing here or why this is marked `PENDING`? I actually haven't seen the `pending` file before, I've always just seen `test`, so I might totally be misunderstanding something here.


EDIT: Ah after seeing https://github.com/sbt/website/commit/c170298e100f3606376dbcbe268f3f2125417921 I'm assuming I can just rename this back to `test` now.

leifwickland:
Thanks for doing this, @ckipp01!

eed3si9n:
Thanks for the contribution @ckipp01! I'm on vacation and I haven't had chance to look into the details yet.

src/sbt-test/ref/example-library/build.sbt
```diff
@@ -1,10 +1,13 @@
...
   .settings(
     name := "Hello",
-    libraryDependencies += "com.typesafe.play" %% "play-json" % "2.9.2",
-    libraryDependencies += "com.eed3si9n" %% "gigahorse-okhttp" % "0.5.0",
-    libraryDependencies += "org.scalatest" %% "scalatest" % "3.2.7" % Test,
+    libraryDependencies ++= Seq(
+      "com.softwaremill.sttp.client4" %% "core" % "4.0.0-M2",
+      "com.lihaoyi" %% "ujson" % "3.1.2",
+      "org.scalatest" %% "scalatest" % "3.2.16" % Test
+    )
```
eed3si9n: Why not use the actual toolkit?

SethTisue: or, if you want to avoid that for some reason, at least MUnit instead of ScalaTest

ckipp01: I don't have a big reason why I didn't just use the toolkit. Mainly because I thought it was nice to show exactly what we were adding since the guide sort of slowly introduces a testing framework, then the other deps. Then they break the other deps out. So if we're just including the toolkit, then you might lose out on some of that. Although, maybe not. Let me go back through and double-check and if I can still illustrate the same things with it I'll switch. Especially swapping out scalatest for munit.

ckipp01:
Alright, so I switched out the usages of explicit dependencies to instead use the `toolkit` and `toolkit-test`. This should be good for a review now!

eed3si9n (approved):
Thanks @ckipp01!

PR #1154, opened 2023-08-11
refactor(ci): just use built-in sbt caching of setup-java

ckipp01:


.github/workflows/ci.yml
```diff
@@ -17,38 +17,32 @@ jobs:
...
+
     - name: Build and test
       run: |
         sbt -v scalafmtSbtCheck scalafmtCheckAll scripted makeSite
-        rm -rf "$HOME/.ivy2/local" || true
-        rm -r $(find $HOME/.sbt/boot -name "*-SNAPSHOT") || true
-        find $HOME/Library/Caches/Coursier/v1        -name "ivydata-*.properties" -delete || true
-        find $HOME/.ivy2/cache                       -name "ivydata-*.properties" -delete || true
-        find $HOME/.cache/coursier/v1                -name "ivydata-*.properties" -delete || true
-        find $HOME/.sbt                              -name "*.lock"               -delete || true
```
ckipp01, line 35: It wasn't really clear to me why this was needed. Is this actually needed for some reason?

eed3si9n: Probably historical artifact. If the CI passes it should be ok?

eed3si9n (approved):

.github/workflows/ci.yml
ckipp01: > If the CI passes it should be ok?

That's what I thought 😆

PR #1174, opened 2024-01-22
Replace references to Gigahorse library

kluen:
The dependency was replaced in 18f165e.

I thought about calling it by its official name "Scala Toolkit", but I guess it's more helpful if it is consistent with how it is referenced below.


eed3si9n (approved):
Thanks!

kluen:
No worries :) I signed the Scala CLA now. Can you redo the check?

PR #1184, opened 2024-02-01
No need to use lazy val to define a string constant

marconilanna:
Specially when the constant is used immediately after

### DIFF
--- a/src/reference/00-Getting-Started/12-Organizing-Build.md
+++ b/src/reference/00-Getting-Started/12-Organizing-Build.md
@@ -86,7 +86,7 @@ import sbt._
 
 object Dependencies {
   // Versions
-  lazy val akkaVersion = "$example_akka_version$"
+  val akkaVersion = "$example_akka_version$"
 
   // Libraries
   val akkaActor = "com.typesafe.akka" %% "akka-actor" % akkaVersion


src/reference/00-Getting-Started/12-Organizing-Build.md
```diff
@@ -86,7 +86,7 @@ import sbt._
 
 object Dependencies {
   // Versions
-  lazy val akkaVersion = "$example_akka_version$"
+  val akkaVersion = "$example_akka_version$"
```
eed3si9n, line 89: While this is true, I feel like people copy-paste build files and run into variable initialization issues, and when they do they may not know how to fix it if they already aren't familiar with Scala. So for the purpose of documentation we should put more `lazy` to reduce on-boarding pains. And if the reader knows Scala, they can remove `lazy`.

marconilanna: That advice is inconsistent with the rest of the file. Why `lazy val` just the string constant but not the module ID definitions below it?

I think this does more harm than good. In my experience, beginners see examples using `lazy val` and they "learn" that they have to put `lazy val` everywhere, too, even where it is clearly not need. This stimulates the spread of a code smell:

https://twitter.com/djspiewak/status/302489756552536064

https://stackoverflow.com/questions/3041253/whats-the-hidden-cost-of-scalas-lazy-val

marconilanna: Anecdotal evidence: when I tell my team they should not abuse `lazy val`, they point back to this example and say: "but it is in the official documentation!". Takes quite a bit of convincing to make them unlearn the bad habit.

eed3si9n: I 100% agree that it's not free. Neither is my time supporting people confused about early initialization causing null pointer error, so for sbt project putting `lazy` everywhere where ppl can run into issue has been an ok workaround. Re: inconsistencies, we should probably more more.

PR #1202, opened 2024-04-12
update the link for downloading diffutils jar

btrachey:


eed3si9n (approved):
Thanks!

btrachey:
> Thanks!

For sure! Would you be open to a PR with some light re-work of the language on that incremental compilation page? I found myself re-reading a couple sentences a few times because the wording was a little confusing.

PR #1203, opened 2024-04-23
update library dependencies docs.

wenhoujx:
says the title: update library dependencies docs.

### DIFF
--- a/src/reference/guide/08-Library-Dependencies.md
+++ b/src/reference/guide/08-Library-Dependencies.md
@@ -171,10 +171,10 @@ of the dependency is declared in the build, and it satisfies the
 range, then sbt will use the specified version.  Otherwise, Coursier could
 go out to the Internet to find the latest version.  This would result
 to a surprising behavior where the effective version keeps changing
-over time, even though there's a specified version of the library that
-satisfies the range condition.
+over time, even though there's a specified version of the library in the build 
+but it doesn't satisfy the range condition.
 
-Maven version ranges will be replaced with its lower bound if the
+Maven version ranges will be replaced with its lower bound in the
 build so that when a satisfactory version is found in the dependency
 graph it will be used.  You can disable this behavior using the JVM
 flag `-Dsbt.modversionrange=false`.


src/reference/guide/08-Library-Dependencies.md
```diff
@@ -171,10 +171,10 @@ of the dependency is declared in the build, and it satisfies the
 range, then sbt will use the specified version.  Otherwise, Coursier could
 go out to the Internet to find the latest version.  This would result
 to a surprising behavior where the effective version keeps changing
-over time, even though there's a specified version of the library that
-satisfies the range condition.
+over time, even though there's a specified version of the library in the build 
+but it doesn't satisfy the range condition.
```
eed3si9n, line 175: I don't think this change is correct. Here's the surprise in bullet points:
- The build specifies a version number, which satisfies the range condition.
- Given the range condition, Coursier or Ivy would go out to the Internet to find the latest version in the range.
- If there's a new version found from the Internet that satisfies the range condition, Coursier or Ivy would use the newly found version, ignoring what the user has specified.
- This can change the behavior of your software.

wenhoujx: hmm, if "The build specifies a version number, which satisfies the range condition", then the sbt uses the specfied version. 
on the other hand, if the build does NOT specify a version number, Coursier could go to internet and get changing latest version, which could be surprising to the build author. 

i think we should rewrite the doc to remove the `even though` part: 

> Occasionally a Maven "version range" is used to specify a dependency
> (transitive or otherwise), such as `[1.3.0,)`.  If a specific version
> of the dependency is declared in the build, and it satisfies the
> range, then sbt will use the specified version.  Otherwise, Coursier could
> go out to the Internet to find the latest version.  This would result
> to a surprising behavior where the effective version keeps changing
> over time. ~~even though there's a specified version of the library in the build 
> but it doesn't satisfy the range condition.~~
>

wenhoujx: @eed3si9n lmk what do you think. thanks for reviewing.

eed3si9n: Here's my take on the rewrite - https://github.com/sbt/website/pull/1206

wenhoujx: commented there .

eed3si9n:
Closing this in favor of https://github.com/sbt/website/pull/1206

PR #1206, opened 2024-04-24
Rewrite the Ivy revisions section

eed3si9n:
Ref https://github.com/sbt/website/pull/1203

### DIFF
--- a/src/reference/guide/08-Library-Dependencies.md
+++ b/src/reference/guide/08-Library-Dependencies.md
@@ -165,8 +165,8 @@ These dynamic versions might work one day, but break the next, and make the buil
 A Maven version range such as `[1.3.0,)` can be used to specify a dependency,
 and it can appear transitively in existing POM files. When a dependency manager
 like Coursier finds a version range, it will go out to the Internet to find the latest version.
-This could lead to surprising behavior where the effective version changes over time,
-even when there is a specified version of the library within the version range.
+Transitive version ranges could lead to surprising behavior where the effective version changes over time,
+even when there is a specific version of the library within the version range in `build.sbt`.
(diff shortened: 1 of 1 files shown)


src/reference/guide/08-Library-Dependencies.md
```diff
@@ -153,32 +153,26 @@ to ensure binary compatibility.
...
+Cousier and Ivy can select the latest revision of a module according to constraints you specify. 
+Instead of a fixed revision like
+`"1.6.1"`, you specify `"latest.integration"`, `"2.9.+"`, or `"[1.0,)"`. 
+These dynamic versions might work one day, but break the next, and make the build non-repeatable.
+
+A Maven version range such as `[1.3.0,)` can be used to specify a dependency,
+and it can appear transitively in existing POM files. When a dependency manager
+like Coursier finds a version range, it will go out to the Internet to find the latest version.
+This could lead to surprising behavior where the effective version changes over time,
+even when there is a specified version of the library within the version range.
```
wenhoujx: im still confused about the referenced `specified version`, if a range was provided as `version`, where did the `specified version` come from?

eed3si9n: I guess transitively from other subproject or libraries.

wenhoujx: ok, mind to make that into the rewrite? i think that might help others who think like me 😝 
just a nit

wenhoujx (approved):
lgtm, but i don't have power.

src/reference/guide/08-Library-Dependencies.md
eed3si9n: Addressed in 55ef0a5.

PR #1210, opened 2024-05-09
Update `docusaurus.config.js` with corrected `projectName` and GitHub link

kevin-lee:
Update `docusaurus.config.js` with corrected `projectName` and GitHub link

- The documentation site is for `sbt`, so the `projectName` should be `sbt`, not `website`.
- The GitHub link should direct to the `sbt` repository rather than the `website` repository, as the doc site itself is for sbt.

**This is what I think it should be, but I don't know the intention of the current config, so this PR might make no sense.**

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -54,7 +54,7 @@ const config = {
   // GitHub pages deployment config.
   // If you aren't using GitHub pages, you don't need these.
   organizationName: 'sbt', // Usually your GitHub org/user name.
-  projectName: 'website', // Usually your repo name.
+  projectName: 'sbt', // Usually your repo name.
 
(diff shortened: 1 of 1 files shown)


eed3si9n (approved):
Thanks!
I guess having a separate repo is not the usual setup for Docusaurus?

kevin-lee:
Although I often have the doc inside the project itself, I don't think having a separate doc repo is that unusual. 
Having the doc site inside the project itself can have a certain benefit.
e.g.)
If you use mdoc to generate the docs, you can easily make your doc module depend on the project module. However, you can also do it with `libraryDependencies`.

Anyway, as I already said, I don't think having a separate doc repo is unusual. You can keep git history cleaner without doc site in the project.

PR #1212, opened 2024-06-04
Add section on disabling scaladoc

majk-p:
Disabling scaladoc is known to improve build times for large codebases (example: https://github.com/scala/scala3/pull/18157). Multiple projects do it (example https://github.com/scala/scala/blob/2.13.x/build.sbt#L362) but it doesn't seem to be documented. Hopefully that's useful.

Also please let me know if there are other/better options to do it.

### DIFF
--- a/src/reference/04-Howto/11-Howto-Scaladoc.md
+++ b/src/reference/04-Howto/11-Howto-Scaladoc.md
@@ -107,3 +107,16 @@ apiURL := Some(url("https://example.org/api/"))
 
 This information will get included in a property of the published
 `pom.xml`, where it can be automatically consumed by sbt.
+
+
+<a name="disable-scaladoc"></a>
+
+### Disable scaladoc
+
+If you don't need the Scaladocs to be generated, you can empty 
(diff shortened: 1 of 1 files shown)


majk-p:
Doing some more search over Github repositories, I noticed combined use of  `Compile / doc / skip := true` along with emptying `Compile / doc / sources`. Should both be referenced in the docs?

eed3si9n:
I don't know if `Compile / doc / skip` does anything, so I'd recommend against that unless it's proven to be useful.

eed3si9n:
> Disabling scaladoc is known to improve build times for large codebases

What does "build times" mean here? I don't think `compile` or `test` would invoke `doc`. Do you mean for `publishLocal`?

majk-p:
> Do you mean for publishLocal?

Yes, in my quite specific case I need to run `publishLocal` against some of my generated sources to run `scripted` tests. Do you think it's better to reword it to something like `reduce the time consumed by publishing task`?

majk-p:
CLA signed

PR #1222, opened 2024-09-10
First attempt at documenting serverIdleTimeout

lervag:
This is not ready for merge. There are a few questions:

1. Is this the correct place to add this type of reference?
2. Is it the correct way to document it?
3. Should I format it differently?

I would not mind completing this and perhaps try and include a few more settings when I'm at it.

Please see discussion at https://github.com/sbt/sbt/issues/7631.

### DIFF
--- a/src/reference/02-DetailTopics/01-Using-sbt/07-sbt-server.md
+++ b/src/reference/02-DetailTopics/01-Using-sbt/07-sbt-server.md
@@ -15,6 +15,18 @@ in addition to accepting input from the terminal, server also to accepts input f
 This allows multiple clients to connect to a _single session_ of sbt.
 The primary use case we have in mind for the client is tooling integration such as editors and IDEs. See [IDE Integration][IDE] page.
 
+### Configuration
+
+There are several settings that can be used to configure the server.
+The following lists some of these and their default values.
+One may change the settings either per project or by setting values in `~/.sbt/1.0/global.sbt`.
+
+```scala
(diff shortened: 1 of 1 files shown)


eed3si9n:
Yea. I guess we don't yet have user-facing sbt server page, but I think this is fine for now.

src/reference/02-DetailTopics/01-Using-sbt/07-sbt-server.md
```diff
@@ -332,3 +332,16 @@ This will respond back with the result of the action.
   [lsp_initialize]: https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#initialize
   [lsp_initialized]: https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#initialized
   [lsp_publishdiagnosticsparams]: https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#publishdiagnostics-notification
+
+### Configuration
```
eed3si9n: I think this should go immediately after the intro, like at line 17.

eed3si9n:
Also could you target 1.x branch for sbt 1.x documentation plz?

lervag:
I believe I've made the suggested changes.

Should we include more options here before considering a merge? If so, are there any particular ones that seem more important?

eed3si9n:
> Should we include more options here before considering a merge? If so, are there any particular ones that seem more important?

I'd say ship first and we can add more later.

PR #1250, opened 2024-10-29
[1.x] Remove deprecated `IntegrationTest` config from doc

### LINKED ISSUE #7467: Docs still describe IntegrationTest config
Even though the `IntegrationTest` config now is deprecated https://www.scala-sbt.org/1.x/docs/Testing.html#Integration+Tests shows it as the way to do integration tests.

Probably at least drop that section. Even better would be to replace with some recommendations on how to structure project integration test modules, include them in aggregate etc.

Friendseeker:
Closes https://github.com/sbt/sbt/issues/7467


eed3si9n (approved):
Thanks!

PR #1274, opened 2025-01-08
Update GitHub Actions guide to include setup-sbt action

stringbean:
This step is needed now that sbt has been removed from the default Ubuntu & macOS runners.

Also bumps the sbt version in the documentation to 1.10.7.

### DIFF
--- a/src/reference/02-DetailTopics/05-Plugins-and-Best-Practices/04-GitHub-Actions-with-sbt.md
+++ b/src/reference/02-DetailTopics/05-Plugins-and-Best-Practices/04-GitHub-Actions-with-sbt.md
@@ -5,6 +5,7 @@ out: GitHub-Actions-with-sbt.html
   [GA]: https://docs.github.com/en/free-pro-team@latest/actions
   [GA-Reference]: https://docs.github.com/en/free-pro-team@latest/actions/reference
   [Setup-Java]: https://github.com/actions/setup-java
+  [Setup-sbt]: https://github.com/sbt/setup-sbt
   [GA-Matrix]: https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategy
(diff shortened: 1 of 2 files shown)


src/reference/02-DetailTopics/05-Plugins-and-Best-Practices/04-GitHub-Actions-with-sbt.md
```diff
@@ -44,20 +46,30 @@ jobs:
...
-        java-version: 8
-    - name: Build and Test
-      run: sbt -v +test
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Setup JDK
+        uses: actions/setup-java@v4
+        with:
+          distribution: temurin
+          java-version: 11
```
mzuehlke, line 55: I think it is good to specify `sbt` caching: https://github.com/actions/setup-java#caching-packages-dependencies
It's use like this here, e.g.: https://github.com/sbt/sbt-native-packager/blob/main/.github/workflows/validate-pr.yml#L19-L24

stringbean: There is already a [section on caching](https://www.scala-sbt.org/1.x/docs/GitHub-Actions-with-sbt.html#Caching) on that page that talks about adding that exact option.

The whole page could do with a full rewrite to bring it up date but that can be done on the new docs site. For now I wanted to do a quick update so that the examples work.

mzuehlke: Sorry, my bad I was only looking at the diff 🙈

eed3si9n (approved):
Thanks!!

eed3si9n:
@stringbean Could you sign the Scala CLA please?

PR #1284, opened 2025-02-17
Add info about running single scripted test

majk-p:
I noticed that running single scripted test is not documented anywhere, so here's a small PR to explain it


eed3si9n (approved):
Thanks!

PR #1286, opened 2025-03-01
remove obsolete Hello World page

### LINKED ISSUE #1167: Obsolete "hello world" page offers obsolete g8 template
on Discord today, a user reported that they googled "scala sbt hello world" and the top hit was https://www.scala-sbt.org/1.x/docs/Hello.html

which is an outdated page that points to the outdated `sbt/scala-seed.g8` template, which is an old fork that gives you really old sbt and Scala versions

possible resolutions:

* leave the page in place and update the contents
* leave the page in place but replace it with a single sentence directing the user to https://www.scala-sbt.org/1.x/docs/sbt-new-and-Templates.html and/or https://www.scala-sbt.org/1.x/docs/sbt-by-example.html
* replace the page with an HTTP redirect to one of those pages
* remove the page entirely

looking for references to the old template, we see:

```
% git grep sbt/scala-seed
src/reference/99-Archive/02-Hello.md:\$ sbt new sbt/scala-seed.g8
src/reference/es/00-Getting-Started/02-sbt-by-example.md:\$ sbt new sbt/scala-seed.g8
src/reference/ja/00-Getting-Started/02-sbt-by-example.md:\$ sbt new sbt/scala-seed.g8
src/reference/ja/99-Archive/02-Hello.md:\$ sbt new sbt/scala-seed.g8
```

so there's also the translations to worry about.

@eed3si9n I'd be happy to a PR a solution if you give me some guidance on what the appropriate solution is. I'm unsure whether the site infrastructure offers redirects, and I'm uncertain if the page needs to somehow be left in place because it's part of historic sbt 0.13 documentation?

SethTisue:
fixes #1167

I wasn't able to test the redirect part locally, as the site I get at `localhost:3000` doesn't seem to include any redirect support at all. I hope it seems right?

if this is deemed mergeable, would I also need to submit a separate 1.x backport? the "Note to contributors" in README.md doesn't address the relationship of the two branches


eed3si9n:
develop branch contains the landing page and the sbt 2.x docs currently. I can backport these to 1.x.

PR #1304, opened 2025-05-02
fix: css layout for landing page

Zerkath:
Small improvements to the landing page:
- fixes issue where on larger monitors hero banner would align left instead of center
- improve button layout for landing page, aligning to a column & with less chance of buttons going outside of bounds to right
- fix issue where on dark-mode hero title would turn dark, giving insufficient contrast.

# Mobile
- [Mobile - Before](https://github.com/user-attachments/assets/cac63d2a-614a-4f23-a7b9-7649a814433c) 
- [Mobile - After](https://github.com/user-attachments/assets/8d1d0f50-141d-483d-b5eb-bf6b03848201)

# Desktop
- [Desktop - Before](https://github.com/user-attachments/assets/017a3032-a6ac-4560-8420-d78b3dfa80f8)
- [Desktop - After](https://github.com/user-attachments/assets/ba3502e5-72c4-4cb3-b9f5-2c4cc38b15c5)

# Dark-mode
- [Before](https://github.com/user-attachments/assets/928d2278-9587-4ee0-a83d-75b8b4de5947)
- [After](https://github.com/user-attachments/assets/23804e73-0a57-40e0-b753-d4f889f4dc63)

### DIFF
--- a/src/pages/index.module.css
+++ b/src/pages/index.module.css
@@ -4,35 +4,30 @@
  */
 
 header {
-  background-color: #222222;
+  background-color: #222;
   margin-bottom: 5em;
 }
 
 .heroBanner {
-  height: 530px;
-  max-width: 2000px;
-  padding: 4rem 0;
+  padding: 10rem 0;
   text-align: center;
-  position: relative;
-  overflow: hidden;
-  background-image: url("/assets/ryoanji-2000x548.jpg");
-  margin-bottom: 0.1em;
-}
-
-@media screen and (max-width: 996px) {
-  .heroBanner {
-    padding: 2rem;
-  }
+  background: url("/assets/ryoanji-2000x548.jpg") center;
+  background-size: cover;
+  /* Prevent header text from turning dark.
+   * Insufficient contrast due to the background image */
+  --ifm-hero-text-color: #fff;
 }
 
 .buttons {
   display: flex;
   align-items: center;
+  justify-content: center;
   flex-wrap: wrap;
+  gap: 2rem;
 }
 
-@media screen and (min-width: 600px) {
+@media screen and (max-width: 600px) {
   .buttons {
-    justify-content: center;
+    flex-direction: column;
   }
 }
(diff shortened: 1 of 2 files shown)


src/pages/index.module.css
```diff
@@ -4,35 +4,30 @@
...
-  background-image: url("/assets/ryoanji-2000x548.jpg");
-  margin-bottom: 0.1em;
-}
-
-@media screen and (max-width: 996px) {
-  .heroBanner {
-    padding: 2rem;
-  }
+  background: url("/assets/ryoanji-2000x548.jpg") center;
+  background-size: cover;
```
Zerkath: This is option one making the banner fit into a monitor fully. This might look bad on ultra wide as the banner will be zoomed in a lot. If the full banner image should be visible and the sides may have some empty space then you would want `center no-repeat` with background-size removed.

eed3si9n: > This is option one making the banner fit into a monitor fully. This might look bad on ultra wide as the banner will be zoomed in a lot.

I'd like to avoid the situation where the resolution gradually degrades over time as people start using wider and wider monitor.

Zerkath: Thanks I will change it to be a set size aligned to the center.

Zerkath: If screen size exceeds the width of image it will look something like this:
image

The question is do we want to have those outside areas with that dark background?
Also what do when the mode is in dark
image

Zerkath: @eed3si9n let me know if you have a preference here and I will include it in the next commit.

eed3si9n: > The question is do we want to have those outside areas with that dark background?

Yea. I think it's ok. It's like you're seeing pillars basically.

> Also what do when the mode is in dark

For now nothing different than the light version is fine, I think.

Zerkath:
You can retry, I have signed the CLA.

eed3si9n (approved):
Thanks!

PR #1430, opened 2026-03-10
Update plugin migration guide

anatoliykmetyuk:
- Added an explicit section on caching and an easy way to migrate tasks with `Def.uncached`.
- Expanded the `target` section to document that `target.value` points to different directories between sbt 1.x and 2.x
- Rewrote PluginCompat section and Migrating Classpath subsection to use sbt2-compat


eed3si9n (approved):
Thanks!

PR #1491, opened 2026-07-10
Add a comment in build.sbt

eed3si9n:


mkurz:
why?

eed3si9n:
https://www.scala-sbt.org/download/ still says 2.0.0 but locally running Eleventy shows

Download___sbt

so I'm trying to force a new deployment.

eed3si9n:
Still not working tho - https://github.com/sbt/sbt.github.com/blob/master/download/index.html

PR #1528, opened 2026-08-22
Update global_base to "$HOME/.config/sbt/2"

takezoe:
The default global base seems to be `$HOME/.config/sbt/2` in sbt 2, but `global_base` still points to `$HOME/.sbt/2`.


eed3si9n (approved):
Thanks!

